PR #18, opened 2019-03-05
Fix/cpplint warnings

TimBuckers:
added roslint to package
- remove the rand() method (not thread safe)
- create array with const variable for size
- don't use strcpy and strcat 

run `catkin_make roslint_march_fake_sensor_data` to verify that there are no warnings


Ishadijcks (approved):
LGTM

PR #26, opened 2019-03-28
Easily switch between march3 and testjoint utilizing the hardware_builder and dynamic hardware_interface_node

Ishadijcks:
Configuring the hardware interface to work with the M3 exoskeleton.

Tried and tested trajectorycontroller on all joints with speed and velocity!

Closes https://github.com/project-march/march-iv/issues/160
Closes https://github.com/project-march/march-iv/issues/150
Closes https://github.com/project-march/march-iv/issues/149


march_hardware/src/March4.cpp
```diff
@@ -15,18 +15,40 @@ namespace march4cpp
...
+
+  Encoder RAJenc = Encoder(12, 1086, 1490, 1301, 0.005);
+  Encoder LAJenc = Encoder(12, 631, 1022, 918, 0.005);
+
+  IMotionCube LHJimc = IMotionCube(3, LHJenc);
+  IMotionCube LKJimc = IMotionCube(5, LKJenc);
+  IMotionCube LAJimc = IMotionCube(7, LAJenc);
+  IMotionCube RHJimc = IMotionCube(8, RHJenc);
+  IMotionCube RKJimc = IMotionCube(10, RKJenc);
+  IMotionCube RAJimc = IMotionCube(12, RAJenc);
```
JitskedeVries: All encoders and IMC's are for joints right? We could loose the J in each name I think

march_hardware/src/March4.cpp
```diff
@@ -15,18 +15,40 @@ namespace march4cpp
...
-  TemperatureGES temperatureGES = TemperatureGES(1, 0);
-  // TODO(Isha, Martijn) double-check these numbers.
-  Encoder enc = Encoder(16, 37961, 59649, 39717, 0.3);
-  IMotionCube imc = IMotionCube(2, enc);
-  // For now we are only running on the test joint
-  // TODO(Isha) refactor this that you can switch between test_joint, MARCH4 (and MARCH3)
-  Joint temp = Joint("test_joint", temperatureGES, imc);
+  Encoder RHJenc = Encoder(16, 22134, 43436, 24515, 0.05);
+  Encoder LHJenc = Encoder(16, 9746, 31557, 11830, 0.05);
+
```
JitskedeVries: What does each number mean and why do they need to be this exact number? Same question for the IMC numbers

march_hardware_interface/config/march-iii/controllers.yaml
```diff
@@ -3,12 +3,17 @@ march:
     state:
       type: joint_state_controller/JointStateController
       publish_rate: 50
+    position:
+      left_hip_joint:
+        type: position_controllers/JointPositionController
+        joint: left_hip
```
JitskedeVries: Why is there only a position controller for one joint?

march_hardware_interface/config/march-iii/joint_limits.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+  left_knee:
+    has_position_limits: true
+    min_position: -0.2
+    max_position: 2
+    has_velocity_limits: true
+    max_velocity: 0.2
+
+    has_soft_limits: true
+    soft_lower_limit: -0.25
```
JitskedeVries: Soft lower limit is further than the position limit

march_hardware_interface/config/march-iii/joint_limits.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+  right_knee:
+    has_position_limits: true
+    min_position: -0.2
+    max_position: 2
+    has_velocity_limits: true
+    max_velocity: 0.2
+
+    has_soft_limits: true
+    soft_lower_limit: -0.25
```
JitskedeVries: Soft lower limit is further than the position limit

march_hardware_interface/config/march-iii/joint_limits.yaml
```diff
@@ -0,0 +1,79 @@
...
+    max_position: 0.296
+    has_velocity_limits: true
+    max_velocity: 0.2
+
+    has_soft_limits: true
+    soft_lower_limit: -0.35
+    soft_upper_limit: 0.293
+    k_position: 10.0
+    k_velocity: 10.0
+
```
JitskedeVries: How are these values determined? Is that documented somewhere? Or will they be revised in the future?

JitskedeVries (changes requested):
Seems okay, especially since it has worked on the exoskeleton. However the some soft joint lower limits are wrong and some things need to be explained.
I did not run the code on my computer, since it must work in combination with the exoskeleton.

march_hardware/src/March4.cpp
Ishadijcks: Check the constructor, will become obsolete by march_hardware_builder.

march_hardware/src/March4.cpp
Ishadijcks: Will become obsolete by march_hardware_builder.

march_hardware_interface/config/march-iii/controllers.yaml
Ishadijcks: Can be removed

march_hardware_interface/config/march-iii/joint_limits.yaml
Ishadijcks: Completely made up, will be refactored and tweaked after march_hardware_builder is merged.

What about it would you like to see documented and why?

JitskedeVries: If they are made up, I think this must be mentioned somewhere. Since if we forget to adjust it, M5 will think that we gave these variables these values for a reason.

JitskedeVries: If we had a reason, I think this must be documented with arguments on the Confluence, since it is an important decision that M5 might want to revise.

JitskedeVries: For example it is important why M3 choose 5 degrees for their software limits, because if we want to change that we should have a reason that is better then their reasoning.

Ishadijcks: Will be done with the final values

JitskedeVries (approved):

PR #32, opened 2019-04-05
Fix empty MarchRobot by constructing it via the initializer list

Ishadijcks:
Fix a bug where the hardware_interface would refer to an empty object instead of the one created

### DIFF
--- a/march_hardware/src/MarchRobot.cpp
+++ b/march_hardware/src/MarchRobot.cpp
@@ -14,14 +14,10 @@
 
 namespace march4cpp
 {
-MarchRobot::MarchRobot()
-{
-}
-
 MarchRobot::MarchRobot(::std::vector<Joint> jointList, ::std::string ifName, int ecatCycleTime)
 {
-  this->jointList = jointList;
-  ethercatMaster.reset(new EthercatMaster(&jointList, ifName, this->getMaxSlaveIndex(), ecatCycleTime));
+  this->jointList = std::move(jointList);
+  ethercatMaster.reset(new EthercatMaster(&this->jointList, ifName, this->getMaxSlaveIndex(), ecatCycleTime));
 }
 
 void MarchRobot::startEtherCAT()
--- a/march_hardware_builder/test/TestAllowedRobots.cpp
+++ b/march_hardware_builder/test/TestAllowedRobots.cpp
@@ -15,11 +15,15 @@ class AllowedRobotTest : public ::testing::Test
 protected:
 };
 
-TEST_F(AllowedRobotTest, TestMarch3)
+TEST_F(AllowedRobotTest, TestMarch3Creation)
 {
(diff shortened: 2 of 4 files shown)


march_hardware/src/MarchRobot.cpp
```diff
@@ -14,14 +14,10 @@
...
 {
-MarchRobot::MarchRobot()
-{
-}
-
 MarchRobot::MarchRobot(::std::vector<Joint> jointList, ::std::string ifName, int ecatCycleTime)
 {
-  this->jointList = jointList;
-  ethercatMaster.reset(new EthercatMaster(&jointList, ifName, this->getMaxSlaveIndex(), ecatCycleTime));
+  this->jointList = std::move(jointList);
```
martijnvandermarel, line 19: Is this still necessary?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -17,10 +17,9 @@ using joint_limits_interface::PositionJointSoftLimitsInterface;
 
 namespace march_hardware_interface
 {
-MarchHardwareInterface::MarchHardwareInterface(ros::NodeHandle& nh, AllowedRobot robotName) : nh_(nh)
+MarchHardwareInterface::MarchHardwareInterface(ros::NodeHandle& nh, AllowedRobot robotName)
+  : nh_(nh), marchRobot(HardwareBuilder(robotName).createMarchRobot())
```
martijnvandermarel, line 21: Nice!

march_hardware_builder/test/TestAllowedRobots.cpp
```diff
@@ -15,11 +15,15 @@ class AllowedRobotTest : public ::testing::Test
 protected:
 };
 
-TEST_F(AllowedRobotTest, TestMarch3)
+TEST_F(AllowedRobotTest, TestMarch3Creation)
 {
   HardwareBuilder hardwareBuilder = HardwareBuilder(AllowedRobot::march3);
-  march4cpp::MarchRobot march3;
-  ASSERT_NO_THROW(march3 = hardwareBuilder.createMarchRobot());
+  ASSERT_NO_THROW(march4cpp::MarchRobot march3 = hardwareBuilder.createMarchRobot());
```
martijnvandermarel, line 21: Change so it doesn't fail because of lack of empty constructor

martijnvandermarel (commented):
I will approve once the test is fixed (edit: seems to be fixed now)

martijnvandermarel (approved):
Good fix

march_hardware/src/MarchRobot.cpp
Ishadijcks: Necessary no, but since the old variable is no longer needed we might as well move it instead of copying.

sopje (approved):
March_hardware_builder test passed

PR #33, opened 2019-04-05
Toggle joint actuation from .yaml file

Ishadijcks:
It is now possible to toggle whether a joint should be actuated or not from the `.yaml` file.
This will make it much easier to calibrate the encoders and create passive joints.

It is no longer required to comment and recompile the `imc.goToOperationEnabled()` function :)

I have updated the old tests and included some new ones to properly test this behaviour.

### DIFF
--- a/march_hardware/include/march_hardware/Joint.h
+++ b/march_hardware/include/march_hardware/Joint.h
@@ -12,7 +12,7 @@ class Joint
 {
 private:
   std::string name;
-  bool actuate;
+  bool allowActuation;
   IMotionCube iMotionCube;
   TemperatureGES temperatureGES;
 
@@ -41,7 +41,7 @@ public:
   friend bool operator==(const Joint& lhs, const Joint& rhs)
   {
     return lhs.name == rhs.name && lhs.iMotionCube == rhs.iMotionCube && lhs.temperatureGES == rhs.temperatureGES &&
-           lhs.actuate == rhs.actuate;
+           lhs.allowActuation == rhs.allowActuation;
   }
 
   friend bool operator!=(const Joint& lhs, const Joint& rhs)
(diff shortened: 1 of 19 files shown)


TimBuckers:
Actuate method should also contain a check if the joint can actuate (also write a test for this)

march_hardware/include/march_hardware/Joint.h
```diff
@@ -12,14 +12,15 @@ class Joint
 {
 private:
   std::string name;
+  bool actuate;
```
martijnvandermarel: Suggestion: change variable name to actuateEnabled, which describes it better imo

martijnvandermarel (commented):
Looks good, but consider my suggestion on variable name change.

march_hardware/include/march_hardware/Joint.h
Ishadijcks: Yeah I thought about this. Something like `canActuate` or `allowActuation` or `actuatePossible` are also options. I went with actuate because it looks nice in the yaml `actuate: true` but it can be changed for sure.

Ishadijcks: @TimBuckers thoughts?

TimBuckers: allowActuation or actuateEnable is a good solution I think

JitskedeVries: @Ishadijcks I believe in the variable name allowActuation, should we do that?

Ishadijcks: Sure

Ishadijcks:
If you change the name, don't forget to update all parameters and tests

JitskedeVries (approved):
The variable actuate is now changed into allowActuation. It builds and it runs!

PR #36, opened 2019-04-18
Print joint positions on startup

Ishadijcks:


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -54,6 +54,13 @@ void MarchHardwareInterface::init()
   joint_velocity_command_.resize(num_joints_);
   joint_effort_command_.resize(num_joints_);
 
+  // Print all joint positions on startup in case initialization fails.
+  for (int i = 0; i < num_joints_; ++i)
+  {
+      this->read();
```
Bart-van-Ingen: Does this read() have to be in the for loop? it doesnt seem to be using the index that the for loop is generating ...

Ishadijcks: Correct, fixed

TimBuckers (approved):

PR #45, opened 2019-05-10
Use new catkin build travis process

Ishadijcks:


build.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
TimBuckers, line 1: This script has as output: "Build succeeded with warnings", which is followed by "Tests failed". 

(As you already mentioned) The output log/terminal is very hard to read. Is this possible to fix? Some warnings are red some are yellow for example.

TimBuckers (changes requested):

build.sh
Ishadijcks: Locally or on travis?

TimBuckers: If you run from workspace it works fine

TimBuckers (approved):

PR #47, opened 2019-05-15
Feature/pdb hardware

### LINKED ISSUE #49: Control joint without PDB
When enabling high voltage via PDB, we should also make a case to run on a test joint without pdb.

TimBuckers:
Pdb hardware interface. All reads are published, most writes can be used/triggered by publishing ros messages.  For now the writes are exposed to make everything easier tot test and validate.

Read Topic:
- /march/pdb/

Write Topics:
- /shutdown_allowed
- /low_voltage_net/on_or_off
- /high_voltage_net/on_or_off\

closes #49


march_hardware/include/march_hardware/HighVoltage.h
```diff
@@ -0,0 +1,50 @@
+// Copyright 2019 Project March.
+#ifndef MARCH_FAKE_SENSOR_DATA_HIGHVOLTAGE_H
+#define MARCH_FAKE_SENSOR_DATA_HIGHVOLTAGE_H
```
martijnvandermarel: Why the FAKE_SENSOR_DATA? Looks like a copypaste error?

march_hardware/include/march_hardware/LowVoltage.h
```diff
@@ -0,0 +1,47 @@
+// Copyright 2019 Project March.
+#ifndef MARCH_FAKE_SENSOR_DATA_LOWVOLTAGE_H
+#define MARCH_FAKE_SENSOR_DATA_LOWVOLTAGE_H
```
martijnvandermarel: Again, should be renamed.

march_hardware/src/HighVoltage.cpp
```diff
@@ -0,0 +1,120 @@
...
+      static_cast<uint16>(this->slaveIndex), static_cast<uint8>(this->netMonitoringOffsets.getHighVoltageEnabled()));
+  return highVoltageEnabled.ui;
+}
+
+void HighVoltage::setNetOnOff(bool on, int netNumber)
+{
+  if (netNumber < 1 || netNumber > 8)
+  {
+    ROS_ERROR_THROTTLE(2, "Can't turn high voltage net %d on, there are only 8 high voltage nets", netNumber);
+    throw std::invalid_argument("Only high voltage net 1 and 8 exist");
```
martijnvandermarel: This is worded wrong. Should be between 1 and 8.

march_hardware/src/HighVoltage.cpp
```diff
@@ -0,0 +1,120 @@
...
+  uint8 currentStateHighVoltageNets = getNetsOperational();
+  bit8 highVoltageNets;
+  highVoltageNets.ui = 1 << (netNumber - 1);
+  if (on)
+  {
+    highVoltageNets.ui |= currentStateHighVoltageNets;
+  }
+  else
+  {
+    // This code is needed when this method is allowed to turn off high voltage
```
martijnvandermarel: Comment can be removed

martijnvandermarel: Maybe replace with comment: // Force bit of the respective net to zero.
Also for the if above: // Force bit of the respective net to one.

march_hardware/src/HighVoltage.cpp
```diff
@@ -0,0 +1,120 @@
...
+  }
+  set_output_bit8(static_cast<uint16>(this->slaveIndex),
+                  static_cast<uint8>(this->netDriverOffsets.getHighVoltageNetOnOff()), highVoltageNets);
+}
+
+void HighVoltage::setHighVoltageOnOff(bool on)
+{
+  if (on && getHighVoltageEnabled())
+  {
+    ROS_ERROR_THROTTLE(2, "All-High-Voltage already enabled");
```
martijnvandermarel: Still not convinced on wording of All-High-Voltage. My suggestion would be to just have "High voltage already enabled". The main issue is that we want to differentiate between turning on/off high voltage via the SSR and via turning on/off a single net. I would suggest a difference in wording where for the first method (via SSR) the focus is on "**enable/disable** high voltage", whereas for the second method (single net) the focus is on "**turning on/off** high voltage net".
Maybe other reviewers can come in with opinions?

martijnvandermarel: The suggested wording above would probably also mean renaming this function, I guess.

march_hardware/src/IMotionCube.cpp
```diff
@@ -63,11 +63,12 @@ void IMotionCube::validateMosiPDOs()
 // Set configuration parameters to the IMC
 void IMotionCube::writeInitialSettings(uint8 ecatCycleTime)
 {
+  ROS_INFO("IMotionCube::writeInitialSettings");
   bool success = true;
   // sdo_bit32(slaveIndex, address, subindex, value);
   // mode of operation
   success &= sdo_bit8(slaveIndex, 0x6060, 0, 8);
-
+  ROS_INFO_STREAM(success);
```
martijnvandermarel: Can be removed?

march_hardware/src/LowVoltage.cpp
```diff
@@ -0,0 +1,86 @@
...
+  // The last bit of the 8 bits represents net 1
+  // The second to last bit of the 8 bits represents net 2
+  return ((operational.ui >> (netNumber - 1)) & 1);
+}
+
+void LowVoltage::setNetOnOff(bool on, int netNumber)
+{
+  if (netNumber == 1)
+  {
+    ROS_ERROR_THROTTLE(2, "Can't turn low voltage net 1 on or off, could cause master suicide");
```
martijnvandermarel: As we discussed, after some thought all ethercat communication would fail when net 2 is turned off, since all slaves in the chain apart from master and pdb are powered by net 2. This would mean that the master could not turn net 2 back on again. Therefore it makes no sense to let the master turn on/off low voltage net 2.
Should this also be reflected in the code here?

march_hardware/src/PowerDistributionBoard.cpp
```diff
@@ -0,0 +1,61 @@
...
+  union bit32 current =
+      get_input_bit32(static_cast<uint16>(this->slaveIndex),
+                      static_cast<uint8>(this->netMonitoringOffsets.getPowerDistributionBoardCurrent()));
+  return current.f;
+}
+
+void PowerDistributionBoard::setMasterOnline()
+{
+  bit8 isOkBit;
+  this->masterOnlineToggle = !this->masterOnlineToggle;
```
martijnvandermarel, line 31: Maybe add a comment explaining what this does / why we do it?

march_hardware/src/PowerDistributionBoard.cpp
```diff
@@ -0,0 +1,61 @@
...
+{
+  bit8 isAllowedBit;
+  isAllowedBit.ui = static_cast<uint8>(isAllowed);
+  set_output_bit8(static_cast<uint16>(this->slaveIndex),
+                  static_cast<uint8>(this->bootShutdownOffsets.getShutdownAllowedByteOffset()), isAllowedBit);
+}
+
+bool PowerDistributionBoard::getMasterShutdownRequested()
+{
+  union bit8 current = get_input_bit8(static_cast<uint16>(this->slaveIndex),
```
martijnvandermarel: current is a bad name for this variable. Name it masterShutdownRequested or something

martijnvandermarel (commented):
Reviewed only files in march_hardware for now.

Ishadijcks (commented):
I agree with @martijnvandermarel on a lot of comments. I'm afraid that I am not familiar enough with the PDB and its internal workings to contribute proper feedback on this pr.

It would have been better as multiple prs for the controller and the hardware changes though.

martijnvandermarel (approved):

PR #55, opened 2019-06-07
WIP: Refactor/pdomap no sorting

martijnvandermarel:
This refactor gives the developer more freedom and therefore also responsibility when mapping the PDOs of the iMotionCubes. 
New method of mapping is still via the addObject method, but additional argument is needed which indicates in which register the object needs to be placed.
Order in which objects are added is now also important, as no sorting is done any more.
Now also ROS_WARNs when sdo message fails.

Still needs to be tested on hardware before merging. Feel free to test yourself.


TimBuckers:
The pdo mapping is probably not the solution for the encoder reset bug

Issue #31, opened 2019-04-03
check for presence of slaves every ethercat cycle


martijnvandermarel:
be able to detect if slaves are lost and notify developer if slave is permanently lost. If temporarily lost, try to re-establish connection?

Ishadijcks:
Can we actively re-establish the connection or is it mostly hoping it works again?
I'm not sure of the frequency of this happening but isn't this definitely a fatal exception?

How does an IMC react if it no longer gets commands? Default to 0 or repeat the last command?

Ishadijcks:
Implemented in #61

PR #61, opened 2019-06-18
Feature/monitor slave connection

### LINKED ISSUE #56: Make sure every EtherCAT cycle has constant time
The amount of time that needs to be slept every ethercat cycle should ideally be adapted based on how long the first part of the ethercat infinite loop takes. This will ensure every ethercat cycle is exactly x microseconds instead of x + time it takes to read/write ethercat buffers + time it takes to monitor slave connections etc.
This can probably be done with the standard chrono boost library in cpp? Time how long the first part of the loop takes (e.g. y microseconds) and then sleep for x-y microseconds. The total loop will then take x microseconds, as desired.

martijnvandermarel:
Closes #56 and #31 

Detects and throws exception when slave is lost (e.g. ethercat cable is pulled loose or device in chain loses power).
Also makes sure every ethercat cycle is constant time. Fixed both because I needed to see how long the monitoring of slaves took every cycle.

Still need help because roslint fails because "chrono is an unapproved C++11 header". Anyone know how to bypass this?


Ishadijcks:
From the google style guide
image

Ishadijcks:
Why so many whitespace changes? Please try to avoid cluttering the pull request.

We throw an exception the first cycle a slave is lost, do we want some buffer here? Has this been tested with real slaves?

A warning is thrown when we don't reach the cycle. Have you encountered this already?
Maybe we need to perform some benchmarking tests to see how fast we can go?

martijnvandermarel:
@Ishadijcks 
> Why so many whitespace changes? Please try to avoid cluttering the pull request.
>
Issues with my autoformatting. Will fix this.
> We throw an exception the first cycle a slave is lost, do we want some buffer here? Has this been tested with real slaves?
>
I have tested this with a single slave. I have only seen the exception being thrown when I clearly removed the ethercat cable, never otherwise. So I don't think a buffer is necessary. Test with more slaves would be nice, though.
> A warning is thrown when we don't reach the cycle. Have you encountered this already?
> Maybe we need to perform some benchmarking tests to see how fast we can go?
>
I already did some benchmarking on my pc. Average time sending/receiving ethercat messages is 130 microseconds, while checking slave connection takes another 110 microseconds on average. So generally no timing issues with our typical ethercat rate of 1-4 ms. Again, only a single slave and not on the panda, so more tests would be nice. Also there seemed to be some outliers, because with an ethercat cycle time of 1-4 ms, still approx. once every minute the warning was thrown. I don't think this is really a big issue, though.

Ishadijcks (approved):
I've fixed the whitespace changes and it's good to merge now!

We may want to test it on the exoskeleton first though.

martijnvandermarel:
I have tested this on the exoskeleton. Correctly detected if for example the fifth slave was lost by pulling out the ethernet cable. 
Constant loop time also works on the exoskeleton, with similar results compared to with my pc.

I think this is ready to be merged if the Travis build passes

PR #68, opened 2019-06-20
Load robot specifications from URDF instead of config yamls

Ishadijcks:
## Introduction
This pr should remove some redundancy in tweaking the exoskeleton. A joint is ignored if it's fixed in the URDF. Instead of having to comment it out in the .yaml

Depends on https://github.com/project-march/march-iv/pull/241 for the correct URDF specs.

Closes #68 

## Features
Load used joints from urdf instead of hardware.yaml.
Load safetylimits from urdf instead of the parameter server.


march_hardware_interface/launch/march4.launch
```diff
@@ -1,7 +1,5 @@
 <launch>
-    <rosparam file="$(find march_hardware_interface)/config/march4/hardware.yaml" command="load"/>
     <rosparam file="$(find march_hardware_interface)/config/march4/controllers.yaml" command="load"/>
```
TimBuckers, line 2: do you need to change controllers everytime you change a joint in the urdf (to fixed or back to revol revolute)?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -111,21 +126,21 @@ void MarchHardwareInterface::init()
...
-    {
-      ROS_FATAL("Joint %s is outside of its softLimits (%f, %f). Actual position: %f", joint_names_[i].c_str(),
-                softLimits.min_position, softLimits.max_position, joint_position_[i]);
-
-      std::ostringstream errorStream;
-      errorStream << "Joint " << joint_names_[i].c_str() << " is out of its softLimits (" << softLimits.min_position
-                  << ", " << softLimits.max_position << "). Actual position: " << joint_position_[i];
-      throw ::std::invalid_argument(errorStream.str());
-    }
+//    this->read();
```
TimBuckers: commented code?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -33,16 +35,29 @@ MarchHardwareInterface::~MarchHardwareInterface() = default;
...
   }
 
-  // Get joint names
-  nh_.getParam("/march/hardware_interface/joints", joint_names_);
+  // Get joint names from urdf
+  for (auto const& urdfJoint : model.joints_)
+  {
+      if (urdfJoint.second->type != urdf::Joint::FIXED)
+      {
+          joint_names_.push_back(urdfJoint.first);
```
TimBuckers, line 52: what if a joint is `allowActuation: true`, but the joint is fixed? This means the joint cannot move but high voltage is set to on for that joint?

TimBuckers (commented):
2 of the 3 comments are very general, about how we can improve this part of the system. However, it's possible that the comments should't be fixed in this pr.

march_hardware_interface/src/march_hardware_interface.cpp
TimBuckers: This is also causing a merge conflict

march_hardware_interface/launch/march4.launch
Ishadijcks: Yes

march_hardware_interface/src/march_hardware_interface.cpp
Ishadijcks: My bad, fixed

TimBuckers (approved):

PR #72, opened 2019-06-27
Fix/pdb startup

TimBuckers:
- prepare actuation also without pdb
- clean up code with haspdb boolean
- turn on high voltage on joints not hardcoded for all nets 
- test launch settings for pdb only


march_hardware_builder/src/robots/pdb.yaml
```diff
@@ -0,0 +1,22 @@
...
+      lowVoltageNet2Current: 13
+      highVoltageNetCurrent: 17
+      lowVoltageState: 3
+      highVoltageOvercurrentTrigger: 2
+      highVoltageEnabled: 1
+      highVoltageState: 4
+    netDriverByteOffsets:
+      lowVoltageNetOnOff: 4
+      highVoltageNetOnOff: 3
+      allHighVoltageOnOff: 2
```
Ishadijcks:
```suggestion
      allHighVoltageOnOff: 2
      
```

march_hardware_builder/src/robots/pdb.yaml
```diff
@@ -0,0 +1,23 @@
...
+
```
Ishadijcks:
```suggestion
```

march_hardware_interface/launch/pdb.launch
```diff
@@ -0,0 +1,15 @@
+<launch>
+    <rosparam file="$(find march_hardware_interface)/config/test_setup/controllers.yaml" command="load"/>
+
+    <param name="robot_description"
+           command="$(find xacro)/xacro --inorder '$(find march_description)/urdf/test-joint.xacro'" />
```
Ishadijcks: Why load the test-joint xacro for the pdf?

Ishadijcks (commented):
Can you try this on a pdb with joints with @martijnvandermarel ?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -84,11 +84,13 @@ void MarchHardwareInterface::init()
...
 
-  if (power_distribution_board_read_.getSlaveIndex() != -1)
+  hasPowerDistributionBoard = marchRobot.getPowerDistributionBoard()->getSlaveIndex() != -1;
+  if (hasPowerDistributionBoard)
   {
-    for (int i = 1; i <= 8; i++)
+    for (int i = 0; i < num_joints_; i++)
     {
-      power_distribution_board_read_.getHighVoltage().setNetOnOff(false, i);
+      int netNumber = marchRobot.getJoint(joint_names_[i]).getNetNumber();
```
martijnvandermarel, line 91: Add an error message if joint has no netnumber (aka it is -1). Just like a few lines down in this method.

martijnvandermarel (commented):
Please add the check I suggested. I presume it hasn't been tested yet since it involves having a joint, but I suspect the current situation will break when a joints doesn't have a netnumber

Ishadijcks (approved):

PR #80, opened 2019-07-02
Feature/outside limits check

TimBuckers:


march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
```diff
@@ -59,11 +59,13 @@ class MarchHardwareInterface : public march_hardware_interface::MarchHardware {
   bool hasPowerDistributionBoard;
   boost::shared_ptr<controller_manager::ControllerManager> controller_manager_;
   double p_error_, v_error_, e_error_;
+  SoftJointLimits soft_limits_;
 
 private:
   void updatePowerNet();
   void updateHighVoltageEnable();
   void updatePowerDistributionBoard();
+  bool isOutSideLimits(int joint_index);
```
Ishadijcks: Outside is one word

march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
```diff
@@ -59,11 +59,13 @@ class MarchHardwareInterface : public march_hardware_interface::MarchHardware {
   bool hasPowerDistributionBoard;
   boost::shared_ptr<controller_manager::ControllerManager> controller_manager_;
   double p_error_, v_error_, e_error_;
+  SoftJointLimits soft_limits_;
```
Ishadijcks: Aren't these already created and added to a handle?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -195,6 +182,8 @@ void MarchHardwareInterface::read(ros::Duration elapsed_time)
 {
   for (int i = 0; i < num_joints_; i++)
   {
+    isOutSideLimits(i);
```
Ishadijcks: `this.`

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -316,4 +305,25 @@ void MarchHardwareInterface::updatePowerNet()
...
   }
 }
+
+bool MarchHardwareInterface::isOutSideLimits(int joint_index)
+{
+  march4cpp::Joint joint = marchRobot.getJoint(joint_names_[joint_index]);
+  if (joint_position_[joint_index] < soft_limits_.min_position ||
+      joint_position_[joint_index] > soft_limits_.max_position)
+  {
+    ROS_FATAL("Joint %s is outside of its soft_limits_ (%f, %f). Actual position: %f",
```
Ishadijcks: Is a fatal which does not throw an exception really a fatal?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -316,4 +305,25 @@ void MarchHardwareInterface::updatePowerNet()
     }
   }
 }
+
+bool MarchHardwareInterface::isOutSideLimits(int joint_index)
```
Ishadijcks: The prefix `is` implies it returns a boolean. This is not used whenever you call it as a false gets never returned. Maybe rename it? Or throw the exception based on the return value?

Ishadijcks (dismissed):

march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
TimBuckers: yes, but the handle has no get limits

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -65,6 +65,15 @@ void MarchHardwareInterface::init()
   joint_position_command_.resize(num_joints_);
   joint_velocity_command_.resize(num_joints_);
   joint_effort_command_.resize(num_joints_);
+  soft_limits_.resize(num_joints_);
+
+  for (int i = 0; i < num_joints_; ++i)
+  {
+    SoftJointLimits soft_limits;
+    getSoftJointLimits(model.getJoint(joint_names_[i]), soft_limits);
+    ROS_INFO("soft_limits_ (%f, %f).", soft_limits.min_position, soft_limits.max_position);
```
JitskedeVries, line 74: Do you want this as an ROS_INFO? To pretend spam it could also be a ROS_DEBUG

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -316,4 +313,25 @@ void MarchHardwareInterface::updatePowerNet()
...
+                       joint_names_[joint_index].c_str(), soft_limits_[joint_index].min_position,
+                       soft_limits_[joint_index].max_position, joint_position_[joint_index]);
+
+    if (joint.canActuate())
+    {
+      std::ostringstream errorStream;
+      errorStream << "Joint " << joint_names_[joint_index].c_str() << " is out of its soft_limits_ ("
+                  << soft_limits_[joint_index].min_position << ", " << soft_limits_[joint_index].max_position
+                  << "). Actual position: " << joint_position_[joint_index];
+      throw ::std::invalid_argument(errorStream.str());
```
JitskedeVries: I do not entirely understand what is happening if the joint is outside of it's softlimits, while able to actuate. It throws an error, but what is happening then? Will all the motors shut down? Does the entire system shut down?

What I have found:
Error: Definitely a problem that should be investigated. SysAdmin should be notified automatically, but doesn't need to be dragged out of bed.
Fatal/Critical: Overall application or system failure that should be investigated immediately. Yes, wake up the SysAdmin. 

I think that we have to investigate this kind of fault immediately. But you probably have a reasoning behind it. Curious to hear it.

JitskedeVries (commented):
Code looks fine to me. However I am wondering whether this should be an error or a fatal. But you probably thought about that.

march_hardware_interface/src/march_hardware_interface.cpp
TimBuckers: This is in the init method, so this runs only once on start-up.

march_hardware_interface/src/march_hardware_interface.cpp
TimBuckers: When an error is thrown (which is not catched/handled) the node will crash, this means in this case:
- The hardware_interface crashes
- SOEM communication stops
- All IMC's goes to fault state
(If I'm correct)

TimBuckers: I changed the invalid argument to run-time error. This is more appropriate. We are still discussing how to handle hardware_interface crashes. For now the safety node is not responding to this, but in the future we probably want to add a ROS_FATAL in the safety node when the hardware_interface crashes.

JitskedeVries (approved):

PR #87, opened 2019-07-02
improve the EtherCAT cycle time monitor so the print says something more useful

martijnvandermarel:
Prints a percentage of how many cycles were not achieved on time once every 10 seconds instead of just printing it every time it doesn't make it which clutters your terminal

### DIFF
--- a/march_hardware/src/EtherCAT/EthercatMaster.cpp
+++ b/march_hardware/src/EtherCAT/EthercatMaster.cpp
@@ -151,7 +151,14 @@ void EthercatMaster::ethercatLoop()
     if (totalLoops >= 10*rate) // Every 10 seconds
     {
         float rateNotAchievedPercentage = 100*(float(rateNotAchievedCount)/totalLoops);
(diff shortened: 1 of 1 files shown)


march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -138,12 +141,20 @@ void EthercatMaster::ethercatLoop()
...
     }
     else
     {
         usleep(ecatCycleTimems * 1000 - duration.count());
     }
+    totalLoops++;
+    if (totalLoops >= 10*rate) // Every 10 seconds
+    {
+        float rateNotAchievedPercentage = 100*(float(rateNotAchievedCount)/totalLoops);
+        ROS_INFO("EtherCAT rate of %d milliseconds per cycle was not achieved for %f percent of all cycles", ecatCycleTimems, rateNotAchievedPercentage);
```
TimBuckers: which rate is acceptable? Now everything is a ROS_INFO

TimBuckers (changes requested):

TimBuckers (approved):

Ishadijcks (approved):

PR #88, opened 2019-07-03
Fix/reset imotioncubes

TimBuckers:


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -247,6 +249,36 @@ void MarchHardwareInterface::write(ros::Duration elapsed_time)
   }
 }
 
+void MarchHardwareInterface::resetIMotionCubesUntilTheyWork()
```
JitskedeVries, line 252: Very clear function name :+1:

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -247,6 +249,36 @@ void MarchHardwareInterface::write(ros::Duration elapsed_time)
...
+    for (int i = 0; i < num_joints_; ++i)
+    {
+      march4cpp::Joint joint = marchRobot.getJoint(joint_names_[i]);
+      if (joint.getAngleIU() == 0)
+      {
+        ROS_ERROR("Joint %s failed (encoder reset)", joint_names_[i].c_str());
+        encoderSetCorrectly = false;
+      }
+    }
+    for (int i = 0; i < num_joints_; ++i)
```
JitskedeVries: Why do we reset all the IMC's while only one of them displays an angle of 0? Could we implement the IMC reset in the if loop where we check whether the angle is 0?

JitskedeVries: Hmmm... It is probably because you want to restart the ethercat when one of the IMC's has been reset. And you only want to do it when the all the IMC's are reset and not after every one of them.

JitskedeVries: I think that this piece of code could be more efficient, but I do not directly see how. If you are okay with it, I think you could merge it.

JitskedeVries (approved):

march_hardware/include/march_hardware/Joint.h
```diff
@@ -29,10 +29,12 @@ class Joint
 
   void initialize(int ecatCycleTime);
   void prepareActuation();
+  void resetIMotionCube();
```
Ishadijcks, line 33: Add a todo to refactor this to make joint less dependent on knowledge of IMCs

Ishadijcks (changes requested):

Ishadijcks (approved):

PR #90, opened 2019-07-03
Upload joint_names to parameter server after parsing urdf

Ishadijcks:
Needed so the names of joints can be accessed everywhere in ROS without parsing the URDF.


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -37,7 +37,7 @@ MarchHardwareInterface::~MarchHardwareInterface()
 void MarchHardwareInterface::init()
 {
   // Start ethercat cycle in the hardware
-  this->marchRobot.startEtherCAT();
+//  this->marchRobot.startEtherCAT();
```
Ishadijcks:
```suggestion
  this->marchRobot.startEtherCAT();
```

sopje (approved):

PR #92, opened 2019-07-04
Throw exception when rad command is too far from current position

Ishadijcks:
Changed the value from 0.2 to 0.05

Our max speed is 1.7 rad/s, at 100hz this means we need to move 0.017 rad per instruction at most. 0.05 was chosen to have a bit of a buffer.

NOTE: This means that if the controller has an error of more than 0.05 rad it will also trigger this limit. This needs to be tested on the exoskeleton

### DIFF
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -99,7 +99,8 @@ void IMotionCube::actuateRad(float targetRad)
 {
   if (std::abs(targetRad - this->getAngleRad()) > 0.05)
   {
-    ROS_ERROR("Target %f exceeds max difference of 0.05 from current %f for slave %d", targetRad, this->getAngleRad(), this->slaveIndex);
+    ROS_ERROR("Target %f exceeds max difference of 0.05 from current %f for slave %d", targetRad, this->getAngleRad(),
+              this->slaveIndex);
     throw std::runtime_error("Target exceeds max difference of 0.05 from current position");
   }
   this->actuateIU(this->encoder.RadtoIU(targetRad));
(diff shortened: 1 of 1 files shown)


march_hardware/src/IMotionCube.cpp
```diff
@@ -97,10 +97,11 @@ void IMotionCube::writeInitialSettings(uint8 ecatCycleTime)
 
 void IMotionCube::actuateRad(float targetRad)
 {
-  if (std::abs(targetRad - this->getAngleRad()) > 0.2)
+  if (std::abs(targetRad - this->getAngleRad()) > 0.05)
```
JitskedeVries: 0.05 is probably okay, should be tested on the exoskeleton

Ishadijcks:
```suggestion
  if (std::abs(targetRad - this->getAngleRad()) > 0.075)
```

march_hardware/src/IMotionCube.cpp
```diff
@@ -97,10 +97,11 @@ void IMotionCube::writeInitialSettings(uint8 ecatCycleTime)
...
   {
-    ROS_ERROR("Target %f exceeds max difference of 0.2 from current %f", targetRad, this->getAngleRad());
-    return;
+    ROS_ERROR("Target %f exceeds max difference of 0.05 from current %f for slave %d", targetRad, this->getAngleRad(),
```
Ishadijcks:
```suggestion
    ROS_ERROR("Target %f exceeds max difference of 0.075 from current %f for slave %d", targetRad, this->getAngleRad(),
```

march_hardware/src/IMotionCube.cpp
```diff
@@ -97,10 +97,11 @@ void IMotionCube::writeInitialSettings(uint8 ecatCycleTime)
...
+              this->slaveIndex);
+    throw std::runtime_error("Target exceeds max difference of 0.05 from current position");
```
Ishadijcks:
```suggestion
    throw std::runtime_error("Target exceeds max difference of 0.075 from current position");
```

JitskedeVries (approved):

PR #99, opened 2019-07-09
Enlarge error margin

JitskedeVries:
The error margin is enlarged from 0.075 to 0.27. This is done because shutting all IMC's down is less safe than a deviation between the desired and the actual position. 
The value 0.27 is determined by looking at the log files of M3 during the Cybathlon. They use the same controller variables as we do right now. And during a certain gait they had an error of 0.266 rad. 
A plot showing this error and an analysis of our gait files during training 2 can be found here: https://confluence.projectmarch.nl:8443/pages/viewpage.action?spaceKey=41TECH&title=Analysis+desired+and+target+error


TimBuckers:
I merged this in the training/9-july branch. If these values work today I think we can merge

Ishadijcks (approved):

PR #100, opened 2019-07-09
Add rosbag node to launch file

JitskedeVries:
No logging was active when the testjoint started moving very suddenly (maybe an encoder shift?). Thus I could not find the problem. I added the automatic logging to the launch file of the test joint.

I choose to log the /march/controller/trajectory/state since this publishes both the error, desired and actual states.


Ishadijcks (dismissed):
@TimBuckers I don't think this is something we want in develop. Especially since it clashes with the bag from `march_launch`. Seems redundant and should just be added locally in my opinion.

TimBuckers (approved):

TimBuckers:
for the test-setup it's doesn clash with launch

JitskedeVries:
> @TimBuckers I don't think this is something we want in develop. Especially since it clashes with the bag from `march_launch`. Seems redundant and should just be added locally in my opinion.

I agree with Tim, therefore I thought we could merge this in develop.

PR #101, opened 2019-07-10
Add SlaveCount check that returns the amount of slaves

Ishadijcks:


martijnvandermarel (commented):
Looks good in general. I have a few questions.
Is the choice to hardcode the ifname (enp2s0) deliberate? If so, why? Can't it be retrieved from the yaml like usual?
Speaking of using the yaml, you could use it to compare the found slaves with the amount of slaves that should be there. Is that something you'd want to use?

Ishadijcks:
This check simply reports the amount, a potential check can be added here.

https://github.com/project-march/march-iv/pull/271.

I don't think it can be automated as the expected slaves is not strictly the same as the amount of slaves in the yaml, or the highest slave in the yaml.

The yaml could be parsed for the `ifname`, but then this would need to be added in the hardware interface to avoid a cyclic dependency

martijnvandermarel:
> The yaml could be parsed for the `ifname`, but then this would need to be added in the hardware interface to avoid a cyclic dependency

What are the downsides to adding it to the hardware interface?

Ishadijcks:
~I have to do more work~ This is clearly a test of the `march_hardware`, which is why I believe it should be added there.

TimBuckers (approved):

PR #104, opened 2019-07-11
Better prints for going to operation enabled

martijnvandermarel:
reduced print rate from every 0.1 sec to every 0.5 sec
add prints before and after indicating which joint is being turned on
add indentation to make it more clear to which joint all these prints belong

these prints have bothered me for a while, imo this is much better
ofc this is bit personal preference, if you have ideas about it feel free to tell me/change stuff

### DIFF
--- a/march_hardware/src/Joint.cpp
+++ b/march_hardware/src/Joint.cpp
@@ -55,7 +55,9 @@ void Joint::prepareActuation()
 {
   if (this->allowActuation)
   {
+    ROS_INFO("Trying to prepare joint %s for actuation", this->name.c_str());
     this->iMotionCube.goToOperationEnabled();
+    ROS_INFO("Joint %s successfully prepared for actuation", this->name.c_str());
   }
   else
   {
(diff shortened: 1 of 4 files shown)


march_hardware/src/Joint.cpp
```diff
@@ -55,7 +55,9 @@ void Joint::prepareActuation()
 {
   if (this->allowActuation)
   {
+    ROS_INFO("Trying to prepare joint %s for actuation", this->name.c_str());
```
Ishadijcks:
```suggestion
    ROS_INFO("Preparing joint %s for actuation", this->name.c_str());
```

Ishadijcks (commented):
With a pr like this, it is usually a good idea to include a before and after of the logs. This makes it much easier to review what has actually changed.

I think you added the indentation to make it clearer that prints 'belong' to a specific joint, but it's hard to tell without an example.

TimBuckers:
image
looks good

TimBuckers (approved):

Ishadijcks:
@martijnvandermarel That looks very good indeed :)

PR #111, opened 2019-07-18
Add/new haa encoder values

Bart-van-Ingen:
The HAA joints have been mechanically completed and have now been given the appropriate encoder values. They have been added to the controller and march4 .yaml files

### DIFF
--- a/march_hardware_builder/src/robots/march4.yaml
+++ b/march_hardware_builder/src/robots/march4.yaml
@@ -1,41 +1,41 @@
 march4:
   ifName: enp2s0
   ecatCycleTime: 4
-  powerDistributionBoard:
-    slaveIndex: 7
-    bootShutdownOffsets:
-      masterOk: 0
-      shutdown: 0
-      shutdownAllowed: 1
-    netMonitorByteOffsets:
-      powerDistributionBoardCurrent: 5
-      lowVoltageNet1Current: 9
-      lowVoltageNet2Current: 13
-      highVoltageNetCurrent: 17
-      lowVoltageState: 3
-      highVoltageOvercurrentTrigger: 2
-      highVoltageEnabled: 1
-      highVoltageState: 4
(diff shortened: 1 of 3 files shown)


march_hardware_builder/src/robots/march4.yaml
```diff
@@ -21,21 +21,21 @@ march4:
...
+          slaveIndex: 6
+          encoder: {resolution: 12, minPositionIU: 3688, maxPositionIU: 4070, zeroPositionIU: 3858, safetyMarginRad: 0.05}
+
+    - left_hip_aa:
+        allowActuation: true
+        netNumber: 5
+        temperatureges: {slaveIndex: 8, byteOffset: 0}
+        imotioncube:
+          slaveIndex: 9
+          encoder: {resolution: 12, minPositionIU: 1010, maxPositionIU: 1401, zeroPositionIU: 1194, safetyMarginRad: 0.05}
```
Ishadijcks: The resolution of the `right_hip_aa` is 382 IU, while the `left_hip_aa` is 391 IU, is this acceptable?

Ishadijcks: In the same vein is the zero position of `right_hip_aa` 170 IU from its start, and `left_hip_aa` 184 IU

Ishadijcks: Could just be that the `minPositionIU` of the `right_hip_aa` is 10 toO HIGH, or the `left_hip_aa` too low.

Bart-van-Ingen: The min and max position of the joints was determined by pushing the joints to their physical min and max position. The difference in range is therefore caused by the hardware. This difference can therefore not be solved through a software solution.

Ishadijcks:
Please fix the conflict

martijnvandermarel (approved):

PR #115, opened 2019-07-23
Add torque objects to PDO mapping

JitskedeVries:
For the future torque mode, we need to map actualtorque and targettorque in the PDO mapping.


Ishadijcks (commented):
Looks good!

Is everyone alright with `ActualTorque` being compulsory? Makes sense IMO

Bart-van-Ingen:
Yes! as it will allow us to monitor the current being pulled by the motor when moving. Therefore it should be compulsory

Ishadijcks (approved):
Agreed

PR #122, opened 2019-07-29
Change joint limits check to use hard limits instead of soft

### LINKED ISSUE #121: Refactor outsidelimits check to use joint_limits, not soft_limits on startup
As we always home on startup, the exoskeleton can also be withing it's hard position limits on startup.

I believe the `hard_limits_` handle was removed so it needs to be readded from the urdf.

Ishadijcks:
Closes #121 

Needs to be tested on the exoskeleton

### DIFF
--- a/march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
+++ b/march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
@@ -68,6 +68,7 @@ protected:
   RtPublisherPtr imc_state_pub_;
   double p_error_, v_error_, e_error_;
   std::vector<SoftJointLimits> soft_limits_;
+  std::vector<JointLimits> joint_limits_;
 
 private:
   void updatePowerNet();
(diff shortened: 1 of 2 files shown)


martijnvandermarel:
@Ishadijcks I believe the error that is described in issue #121 is thrown in the imotioncube::goToOperationEnabled method (https://github.com/project-march/hardware-interface/blob/develop/march_hardware/src/IMotionCube.cpp), not the outsideLimitsCheck method in the hardware interface. So I believe this PR doesn't fix the issue?

Ishadijcks:
I'm not sure which check you mean, the one I see uses the encoder values, that should be correct?

JitskedeVries:
> @Ishadijcks I believe the error that is described in issue #121 is thrown in the imotioncube::goToOperationEnabled method (https://github.com/project-march/hardware-interface/blob/develop/march_hardware/src/IMotionCube.cpp), not the outsideLimitsCheck method in the hardware interface. So I believe this PR doesn't fix the issue?

Plus I think we should still check whether we are inside the softlimits instead of the jointlimits before we can actuate. It is more safe.

Ishadijcks:
That is the current behaviour.

Ishadijcks:
@martijnvandermarel we can close this pr now right?

martijnvandermarel:
> @martijnvandermarel we can close this pr now right?

Yes I suppose so.

TimBuckers:
Delete this branch?

PR #125, opened 2019-07-30
Add actuation mode to joint and imc

Ishadijcks:
## Features
- Add actuation mode attribute to IMotionCube class
- Allow actuation mode to be set once at runtime
- Add actuation mode of position
- Only allow actuation in radians when in position mode


march_hardware/include/march_hardware/IMotionCube.h
```diff
@@ -64,6 +72,15 @@ class IMotionCube : public Slave
   bool goToOperationEnabled();
   bool resetIMotionCube();
 
+  void setActuationMode(ActuationMode actuationMode)
```
Bart-van-Ingen: why are these methods in the header file and not in .cpp?

Ishadijcks: Laziness probably, fixed

Bart-van-Ingen (approved):

martijnvandermarel (approved):

PR #126, opened 2019-07-30
change min/max pos in encoder class to pos/neg soft- and hardlimits

### LINKED ISSUE #121: Refactor outsidelimits check to use joint_limits, not soft_limits on startup

martijnvandermarel:
Closes #121 

Refactors the Encoder class to have both hardlimits and softlimits. This means that when actuating to a certain position, there is the possibility to check whether the targetposition is within softlimits, while before homing there is also the possibility to check if joints are within hardlimits.

### DIFF
--- a/march_hardware/src/Encoder.cpp
+++ b/march_hardware/src/Encoder.cpp
@@ -17,18 +17,18 @@ Encoder::Encoder(int numberOfBits, int minPositionIU, int maxPositionIU, int zer
 
   this->safetyMarginRad = safetyMarginRad;
   this->slaveIndex = -1;
-  this->positiveHardLimitIU = maxPositionIU;
-  this->negativeHardLimitIU = minPositionIU;
+  this->upperHardLimitIU = maxPositionIU;
+  this->lowerHardLimitIU = minPositionIU;
   this->zeroPositionIU = zeroPositionIU;
   int safetyMarginIU = RadtoIU(safetyMarginRad) - this->zeroPositionIU;
-  this->positiveSoftLimitIU = this->positiveHardLimitIU - safetyMarginIU;
-  this->negativeSoftLimitIU = this->negativeHardLimitIU + safetyMarginIU;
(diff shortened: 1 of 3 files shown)


Ishadijcks (changes requested):
I disagree with the naming of positive and negative. It implies one should be above 0 and the other below. Can you change it to `upper` and `lower`? It's more in line with ROS.

Other than that, good fix :)

march_hardware/src/Encoder.cpp
```diff
@@ -17,18 +17,18 @@ Encoder::Encoder(int numberOfBits, int minPositionIU, int maxPositionIU, int zer
...
-  ROS_ASSERT_MSG(this->minPositionIU < this->maxPositionIU, "Invalid range of motion. Safety margin too large or "
-                                                            "min/max position invalid. Minposition: %i, Maxposition: "
-                                                            "%i, safetyMargin: %f",
-                 this->minPositionIU, this->maxPositionIU, this->safetyMarginRad);
+  this->upperSoftLimitIU = this->upperHardLimitIU - safetyMarginIU;
+  this->lowerSoftLimitIU = this->lowerHardLimitIU + safetyMarginIU;
+
+  ROS_ASSERT_MSG(this->lowerSoftLimitIU < this->upperSoftLimitIU,
+                 "Invalid range of motion. Safety margin too large or "
+                 "min/max position invalid. Minposition: %i IU, Maxposition: "
```
Ishadijcks: Change print to upper and lower soft limits

Ishadijcks: Also print the hard limits

Ishadijcks (changes requested):
One small change

Ishadijcks (approved):
Good fix

PR #127, opened 2019-07-30
Refactor the traversing of the IMC state machine

Ishadijcks:
Should be tested on the exoskeleton/testjoint.

Provides a much cleaner way to switch between states :)

### DIFF
--- a/march_hardware/include/march_hardware/IMotionCubeTargetState.h
+++ b/march_hardware/include/march_hardware/IMotionCubeTargetState.h
@@ -1,0 +1,54 @@
+// Copyright 2019 Project March.
+
+#ifndef MARCH_WS_IMOTIONCUBETARGETSTATE_H
+#define MARCH_WS_IMOTIONCUBETARGETSTATE_H
+
+#include <string>
+namespace march4cpp
+{
+
+class IMotionCubeTargetState
+{
+public:
+  static const IMotionCubeTargetState SWITCH_ON_DISABLED;
+  static const IMotionCubeTargetState READY_TO_SWITCH_ON;
+  static const IMotionCubeTargetState SWITCHED_ON;
+  static const IMotionCubeTargetState OPERATION_ENABLED;
+
+private:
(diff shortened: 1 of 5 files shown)


march_hardware/include/march_hardware/IMotionCubeTargetState.h
```diff
@@ -0,0 +1,54 @@
...
+namespace march4cpp
+{
+
+class IMotionCubeTargetState
+{
+public:
+  static const IMotionCubeTargetState SWITCH_ON_DISABLED;
+  static const IMotionCubeTargetState READY_TO_SWITCH_ON;
+  static const IMotionCubeTargetState SWITCHED_ON;
+  static const IMotionCubeTargetState OPERATION_ENABLED;
```
martijnvandermarel, line 15: Doesn't it cause recursive behavior to have objects as attributes in their own class definition? If it works at all? Seems like it should be outside the class, or I am missing some knowledge on cpp that allows this

martijnvandermarel (commented):
Nice refactor, see comment

march_hardware/include/march_hardware/IMotionCubeTargetState.h
Ishadijcks: I see where you are coming from, but here I only 'declare' them. Basically telling the compiler that they exist. Also note they are static (shared among all instances of the class)

We later instantiate them from the .cpp file, which actually creates the objects with a 1 level recursion indeed. However the recursed objects are not yet instantiated, only declared. So `IMotionCubeTargetState::SWITCH_ON_DISABLED::SWITCH_ON_DISABLED` exists, but is only declared so a null pointer.


Implementation based on this question:
https://stackoverflow.com/questions/1965249/how-to-write-a-java-enum-like-class-with-multiple-data-fields-in-c

Explanation based on my current knowledge of cpp, not guaranteed to be correct :')

JitskedeVries (approved):
Works on test joint

PR #132, opened 2019-07-31
Add actuateTorque method to iMotionCube and Joint

JitskedeVries:
Also some style things are adjusted.

ActuateTorque method is added to iMotionCube and Joint. However it is not yet implemented in the march_hardware_interface.

### DIFF
--- a/march_hardware/include/march_hardware/ActuationMode.h
+++ b/march_hardware/include/march_hardware/ActuationMode.h
@@ -13,6 +13,7 @@ public:
   enum Value : int
   {
     position,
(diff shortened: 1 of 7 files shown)


Ishadijcks (changes requested):
Nice and small pr, it does however need to add some tests.

Please implement at least the following tests:

- ActuateTorque a position joint
- ActuateRad a torque joint
- Switch from position to torque mode
- Switch from torque to position mode

Ishadijcks (approved):

PR #133, opened 2019-07-31
Validate targetUI based on which direction we are moving

### LINKED ISSUE #131: When outside soft limits, allow actuation to within limits
Now an error is thrown after startup when outside softlimits because joints are actuated to current position which isn't allowed when outside softlimits.

Ishadijcks:
Closes #131

Quite an important change, please review it carefully to make sure I did not make a logic error.
I would like to have 3 people reviewing it at least.

@martijnvandermarel can you perform the same check where you encountered this error to ensure it works?

### DIFF
--- a/march_hardware/src/Encoder.cpp
+++ b/march_hardware/src/Encoder.cpp
@@ -78,6 +78,26 @@ bool Encoder::isWithinSoftLimitsIU(int positionIU)
   return (positionIU > this->lowerSoftLimitIU && positionIU < this->upperSoftLimitIU);
 }
 
+bool Encoder::isValidTargetIU(int currentIU, int targetIU)
+{
+  if (this->isWithinSoftLimitsIU(targetIU))
+  {
+    return true;
+  }
+
+  if (currentIU >= this->getUpperSoftLimitIU())
+  {
+    return (targetIU <= currentIU) && (targetIU > this->getLowerSoftLimitIU());
+  }
+
+  if (currentIU <= this->getLowerSoftLimitIU())
+  {
+    return (targetIU >= currentIU) && (targetIU < this->getUpperSoftLimitIU());
+  }
+
+  return false;
+}
+
 int Encoder::getUpperSoftLimitIU() const
 {
   return upperSoftLimitIU;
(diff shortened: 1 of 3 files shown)


martijnvandermarel:
> @martijnvandermarel can you perform the same check where you encountered this error to ensure it works?

In case I don't have time or someone else wants to do this:
- Move one of the joints (almost) against a physical endstop
- Start up the hardware interface
- See if the error thrown in this line occurs: https://github.com/project-march/hardware-interface/blob/6d6801262badfa1b279eba7a4f5a253ea59ab6f9/march_hardware/src/IMotionCube.cpp#L131

On the current develop, this error is thrown. With this branch, it shouldn't.

martijnvandermarel (commented):
Logic checks out. Approval when tested on the exoskeleton.

JitskedeVries (commented):
The logic is good. You can also get my approval if it is tested on the test setup or the exoskeleton.

march_hardware/src/Encoder.cpp
```diff
@@ -78,6 +78,26 @@ bool Encoder::isWithinSoftLimitsIU(int positionIU)
...
+  if (currentIU >= this->getUpperSoftLimitIU())
+  {
+    return (targetIU <= currentIU) && (targetIU > this->getLowerSoftLimitIU());
+  }
+
+  if (currentIU <= this->getLowerSoftLimitIU())
+  {
+    return (targetIU >= currentIU) && (targetIU < this->getUpperSoftLimitIU());
+  }
+
```
Bart-van-Ingen, line 97: im not sure if this is a relevant detail but should the order of these if statements not be different? With the first one actually being at the bottom? In my mind it makes sense to always first check the current IU with respect to the limits and only once it within them to allow for unrestricted movement within soft limits

Bart-van-Ingen (changes requested):

march_hardware/src/Encoder.cpp
Ishadijcks: Why? If the target is within limits, movements is always allowed

Ishadijcks: @Bart-van-Ingen ping

Ishadijcks: @Bart-van-Ingen ping

TimBuckers (approved):

march_hardware/src/Encoder.cpp
Bart-van-Ingen: rereading this I see that it does make sense

Bart-van-Ingen (approved):

PR #136, opened 2019-08-01
Actuate Torque

JitskedeVries:
The last pullrequest for torque mode! When this is added we could move with torque mode on the exoskeleton. The code is tested on the testjoint. And I want to validate it tomorrow on the exoskeleton.

- Initialize goToOperationEnabled method with 0 torque
- Add actuate torque to the march_hardware_interface.cpp
- Add enforce limits for torque mode
- Add handle for effort interface
- Add controllers

### DIFF
--- a/march_hardware_interface/config/march4/controllers.yaml
+++ b/march_hardware_interface/config/march4/controllers.yaml
@@ -11,6 +11,7 @@ march:
       publish_rate: 1
     trajectory:
       type: position_controllers/JointTrajectoryController
+      allow_partial_joints_goal: true
       # When you switch to effort actuationmode, 'position_controllers' must be substituted with 'effort_controllers' and the gains must be added
       joints:
         - right_hip_fe
@@ -26,8 +27,8 @@ march:
 #        left_hip_fe: {p: 80, i: 0, d: 10, i_clamp: 100, publish_state: true, antiwindup: true}
 #        right_knee: {p: 50, i: 0, d: 10, i_clamp: 100, publish_state: true, antiwindup: true}
 #        left_knee: {p: 50, i: 0, d: 10, i_clamp: 100, publish_state: true, antiwindup: true}
-#        right_ankle: {p: 50, i: 50, d: 3, i_clamp: 100, publish_state: true, antiwindup: true}
-#        left_ankle: {p: 50, i: 50, d: 3, i_clamp: 100, publish_state: true, antiwindup: true}
-#        right_hip_aa: {p: 20, i: 0, d: 5, i_clamp: 100, publish_state: true, antiwindup: true} # TODO(BaCo) tune HAA joints
-#        left_hip_aa: {p: 20, i: 0, d: 5, i_clamp: 100, publish_state: true, antiwindup: true}
+#        right_ankle: {p: 140, i: 0, d: 5, i_clamp: 100, publish_state: true, antiwindup: true}
+#        left_ankle: {p: 160, i: 0, d: 6, i_clamp: 100, publish_state: true, antiwindup: true}
+#        right_hip_aa: {p: 340, i: 0, d: 5, i_clamp: 100, publish_state: true, antiwindup: true} # TODO(BaCo) tune HAA joints
+#        left_hip_aa: {p: 340, i: 0, d: 5, i_clamp: 100, publish_state: true, antiwindup: true}
 
--- a/march_hardware_interface/config/march4/controllers.yaml
+++ b/march_hardware_interface/config/march4/controllers.yaml
@@ -32,3 +32,28 @@ march:
 #        right_hip_aa: {p: 340, i: 0, d: 5, i_clamp: 100, publish_state: true, antiwindup: true} # TODO(BaCo) tune HAA joints
 #        left_hip_aa: {p: 340, i: 0, d: 5, i_clamp: 100, publish_state: true, antiwindup: true}
 
+      constraints:
+        right_hip_fe:
+          trajectory: 0.27
+          goal: 0.27
+        left_hip_fe:
+          trajectory: 0.27
+          goal: 0.27
+        right_knee:
+          trajectory: 0.27
+          goal: 0.27
+        left_knee:
+          trajectory: 0.27
+          goal: 0.27
+        right_ankle:
+          trajectory: 0.27
+          goal: 0.27
+        left_ankle:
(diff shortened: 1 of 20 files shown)


march_hardware_interface/config/march4/controllers.yaml
```diff
@@ -20,3 +21,13 @@ march:
         - left_ankle
         - right_hip_aa
         - left_hip_aa
+#      gains: # Required because we're controlling an effort interface
+#        right_hip_aa: {p: 20, i: 0, d: 5, i_clamp: 100, publish_state: true, antiwindup: true} # TODO(BaCo) tune HAA joints
```
martijnvandermarel: I think it's best if the joint order for the gains is the same as the joint order defined a few lines above

march_hardware_interface/config/test_setup/controllers.yaml
```diff
@@ -9,7 +9,14 @@ march:
     pdb_state:
       type: march_pdb_state_controller/MarchPdbStateController
       publish_rate: 50
-    trajectory:
+    trajectory_position:
```
martijnvandermarel: Why trajectory_position here instead of just trajectory? For the march4 this is still trajectory

martijnvandermarel: I get that there is another trajectory_effort part, but why is it different from the march4 controllers.yaml?

march_hardware_interface/launch/testsetup.launch
```diff
@@ -19,7 +19,8 @@
           args="
                 /march/controller/joint_state
                 /march/controller/temperature_sensor
-                /march/controller/trajectory
+                /march/controller/trajectory_position
+                /march/controller/trajectory_effort
```
martijnvandermarel: Why both? Can you use both at the same time, or does one need to be commented?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -142,33 +146,42 @@ void MarchHardwareInterface::init()
...
+      // Create joint effort limit interface
+      EffortJointSoftLimitsHandle jointEffortLimitsHandle(jointEffortHandle, limits, soft_limits_[i]);
+      effortJointSoftLimitsInterface.registerHandle(jointEffortLimitsHandle);
+    }
 
     // Set the first target as the current position
     this->read();
     joint_velocity_[i] = 0;
     joint_effort_[i] = 0;
+    joint_effort_command_[i] = 0;
```
martijnvandermarel: Shouldn't the effort command only be set when actuationmode is torque? Same as position command the line below: shouldn't it be set only when actuationmode is position?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -255,19 +268,31 @@ void MarchHardwareInterface::read(ros::Duration elapsed_time)
 
 void MarchHardwareInterface::write(ros::Duration elapsed_time)
 {
-  positionJointSoftLimitsInterface.enforceLimits(elapsed_time);
-
   for (int i = 0; i < num_joints_; i++)
   {
     march4cpp::Joint joint = marchRobot.getJoint(joint_names_[i]);
 
+    // Enlarge joint_effort_command so dynamic reconfigure can be used inside it's bounds
```
martijnvandermarel: Only when actuationmode = torque?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -255,19 +268,31 @@ void MarchHardwareInterface::read(ros::Duration elapsed_time)
...
+    joint_effort_command_[i] = joint_effort_command_[i] * 1000;
+
+    positionJointSoftLimitsInterface.enforceLimits(elapsed_time);
+    effortJointSoftLimitsInterface.enforceLimits(elapsed_time);
```
martijnvandermarel: Again, dependent on actuationmode?

martijnvandermarel (changes requested):

march_hardware_interface/config/test_setup/controllers.yaml
JitskedeVries: The march4 controllers have to cooperate with the gait schedular, which expects the name trajectory controller. For the test joint it doesn't matter, because the gait schedular is not used. Therefore it is even easier if both controllers are defined, since then only the name of the actuationmode have to be changed in the yaml.

march_hardware_interface/launch/testsetup.launch
JitskedeVries: See comment above

march_hardware_interface/src/march_hardware_interface.cpp
JitskedeVries: It doesn't really matter. The effort or position command is only sent towards the IMC when in that certain mode.

march_hardware_interface/src/march_hardware_interface.cpp
JitskedeVries: I had to set this before the limits were enforced. Because last time the torque mode was reviewed, Isha requested that both limits had to be enforced. I could both move inside the if loop.

march_hardware_interface/config/test_setup/controllers.yaml
JitskedeVries: For uniformity I decided that I'll still change it to look like the march4.yaml.

martijnvandermarel (approved):
Looks good

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -267,7 +285,20 @@ void MarchHardwareInterface::write(ros::Duration elapsed_time)
...
+
+      if (joint.getActuationMode() == march4cpp::ActuationMode::position)
+      {
+        positionJointSoftLimitsInterface.enforceLimits(elapsed_time);
+        joint.actuateRad(static_cast<float>(joint_position_command_[i]));
+      }
+      else if (joint.getActuationMode() == march4cpp::ActuationMode::torque)
+      {
+        // Enlarge joint_effort_command so dynamic reconfigure can be used inside it's bounds
+        joint_effort_command_[i] = joint_effort_command_[i] * 1000;
```
TimBuckers, line 309: Validate if this is working correctly

PR #138, opened 2019-08-06
refactor sdo_bit functions to generic sendSDOMessage function

martijnvandermarel:
Passes length of message as enum argument instead of having 3 different functions

### DIFF
--- a/march_hardware/include/march_hardware/EtherCAT/EthercatSDO.h
+++ b/march_hardware/include/march_hardware/EtherCAT/EthercatSDO.h
@@ -6,9 +6,15 @@
 
 namespace march4cpp
 {
-int sdo_bit8(int slave, uint32_t index, uint8_t sub, uint8_t value);
-int sdo_bit16(int slave, uint32_t index, uint8_t sub, uint16_t value);
-int sdo_bit32(int slave, uint32_t index, uint8_t sub, uint32_t value);
+
+enum class SDOMessageLength
+{
+  bit8,
+  bit16,
+  bit32
+};
+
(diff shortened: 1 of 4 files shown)


march_hardware/include/march_hardware/EtherCAT/EthercatSDO.h
```diff
@@ -6,9 +6,15 @@
...
-int sdo_bit8(int slave, uint32_t index, uint8_t sub, uint8_t value);
-int sdo_bit16(int slave, uint32_t index, uint8_t sub, uint16_t value);
-int sdo_bit32(int slave, uint32_t index, uint8_t sub, uint32_t value);
+
+enum class SDOMessageLength
+{
+  bit8,
+  bit16,
+  bit32
+};
```
Ishadijcks, line 16: If you give this enum a value there is no need to have a switch case

Ishadijcks (commented):
Not sure if this is actually cleaner

march_hardware/include/march_hardware/EtherCAT/EthercatSDO.h
martijnvandermarel: I still need a switch statement for the value address argument (6th argument) of the ec_SDOwrite function. I need to pass different variables (uint8, uint16 or uint32) there based on the length.

martijnvandermarel:
> Not sure if this is actually cleaner

I get what you mean, but I think at least calling this function is more intuitive, plus more up to style guide

martijnvandermarel:
Btw I'm open to closing this and just making a PR with only style fixes / cleanup for this code if you really think it's better the way it was.

Ishadijcks:
Agreed

PR #139, opened 2019-08-08
Refactor hardware_interface to keep checking IMC reset until all are in operation enabled

Ishadijcks:
@martijnvandermarel @Bart-van-Ingen @TimBuckers 
Please check the logic so I can test it on the exoskeleton.

### DIFF
--- a/march_hardware_interface/src/march_hardware_interface.cpp
+++ b/march_hardware_interface/src/march_hardware_interface.cpp
@@ -285,7 +285,8 @@ bool MarchHardwareInterface::prepareJointsForActuation()
         }
       }
       bool prepareActuation = joint.prepareActuation();
-      if (!prepareActuation){
+      if (!prepareActuation)
+      {
         return false;
       }
     }
--- a/march_hardware_interface/src/march_hardware_interface.cpp
+++ b/march_hardware_interface/src/march_hardware_interface.cpp
@@ -87,15 +87,6 @@ void MarchHardwareInterface::init()
     soft_limits_[i] = soft_limits;
   }
 
-  resetIMotionCubesUntilTheyWork();
-
-  // Print all joint positions on startup in case initialization fails.
-  this->read();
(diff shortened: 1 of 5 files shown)


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -173,25 +164,19 @@ void MarchHardwareInterface::init()
...
-          ROS_FATAL("Joint %s has no high voltage net number", joint.getName().c_str());
-          throw std::runtime_error("Joint has no high voltage net number");
-        }
-      }
-      joint.prepareActuation();
-    }
+  this->prepareJointsForActuation();
+  bool encoderSetCorrectly = this->IMotionCubesWorkingCorrectly();
+
+  while (!encoderSetCorrectly)
```
Bart-van-Ingen:
```suggestion
  while (!this->IMotionCubesWorkingCorrectly())
```
an extra boolean is not really needed. You can just use the method directly, right?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -173,25 +164,18 @@ void MarchHardwareInterface::init()
...
-    }
+  bool jointsPrepared = this->prepareJointsForActuation();
+  bool encoderSetCorrectly = this->IMotionCubesWorkingCorrectly();
+
+  while (!encoderSetCorrectly || !jointsPrepared)
+  {
+    ROS_INFO("Resetting IMotionCubes");
+    this->resetIMotionCubes();
+
+    jointsPrepared = this->prepareJointsForActuation();
```
Bart-van-Ingen, line 177: I think that you need to have high voltage on before you reset the imc. They will not reach operation enabled without high voltage

Bart-van-Ingen (dismissed):

Ishadijcks:
Does not fix the problem

PR #147, opened 2019-08-29
Make 2 different error messages for outside limits / encoder reset

### LINKED ISSUE #144: Differentiate error messages when outside limits or encoder reset
Currently, the same error message gets thrown if either the exoskeleton is outside soft limits at startup or when an encoder reset has occurred during startup.
The message is as follows: 
`"Encoder of iMotionCube (with slaveindex %d) is not functioning properly, read value %d, min value is %d, max value is %d. Shutting down"`
From the read value, it can be seen if the encoder is reset (read value 0 or close to 0) or if it is outside soft limits (read value near min or max value). However, it would be better if the error message is different altogether.

martijnvandermarel:
Should make it easier to see if an encoder reset has occured, or if a joint is outside its limits.

closes #144

### DIFF
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -484,7 +484,7 @@ bool IMotionCube::goToOperationEnabled()
                                                                                   "mapping, so can't get angle");
 
   int angleRead = this->encoder.getAngleIU(this->misoByteOffsets[IMCObjectName::ActualPosition]);
-  //  If the encoder is functioning correctly, move the joint to its current
+  //  If the encoder is functioning correctly and the joint is not outside hardlimits, move the joint to its current
   //  position. Otherwise shutdown
   if (this->encoder.isWithinHardLimitsIU(angleRead) && angleRead != 0)
   {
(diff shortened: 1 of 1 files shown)


TimBuckers:
Encoder reset is sometimes also -1 right?

martijnvandermarel:
> Encoder reset is sometimes also -1 right?

Yes. I thought about this but couldn't find a nice way to fix it. -1 (or -2, +1, etc) are not covered by this check anyway, and since for  the LHFE -1 is a valid encoder value, I noticed this check wouldn't catch an encoder reset to -1 of that joint. However, we havent seen it in practice. 

What do you suggest we do? Checking for a range of values (-2 to +2, for example) would be a fix, I guess?

JitskedeVries (commented):
Could we reset the IMC when we notice the encoder reset? 

I agree on the range of -2 to 2 for the encoder reset. Maybe an exception for the LHFE joint? But that could be quite ugly.

martijnvandermarel:
> Could we reset the IMC when we notice the encoder reset?
> 
> I agree on the range of -2 to 2 for the encoder reset. Maybe an exception for the LHFE joint? But that could be quite ugly.

What do you mean with exception?

For any joint except the LHFE, -2 to 2 are already caught since they are outside the hard limits. Exactly in the case of the LHFE, -1 for example is not caught (since it's within hard limits). A ranged check (-2 to 2) could catch this, but on the other hand if the joint is really at -1, an exception is thrown while it wasn't necessary. So increasing that range would not be without cost.

Resetting the IMC when we notice it has been tried by Isha in a past PR, but it didn't work for some reason.

JitskedeVries:
> > Could we reset the IMC when we notice the encoder reset?
> > I agree on the range of -2 to 2 for the encoder reset. Maybe an exception for the LHFE joint? But that could be quite ugly.
> 
> What do you mean with exception?
> 
> For any joint except the LHFE, -2 to 2 are already caught since they are outside the hard limits. Exactly in the case of the LHFE, -1 for example is not caught (since it's within hard limits). A ranged check (-2 to 2) could catch this, but on the other hand if the joint is really at -1, an exception is thrown while it wasn't necessary. So increasing that range would not be without cost.
> 
> Resetting the IMC when we notice it has been tried by Isha in a past PR, but it didn't work for some reason.

At second thought I think the range should not be enlarged. Our successors might run in to this where the LHFE is actually -1. So when we are not mentioning this anywhere, this PR could become a bug for them.

JitskedeVries (approved):

PR #148, opened 2019-08-30
throw fatal if to fault state while going to operation enabled

### LINKED ISSUE #145: Throw error when stuck in 'Waiting for Operation Enabled'
Sometimes the IMC goes to fault state when waiting for the final transition to 'Operation Enabled'. This can be seen from a (sometimes called 'bit flip') change in the printed status word.
It would save time if the fault state would be detected (through the status word change) and an error is thrown with an appropriate message.

martijnvandermarel:
closes #145 

haven't tested this on the exoskeleton yet.

### DIFF
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -467,7 +467,8 @@ bool IMotionCube::goToTargetState(IMotionCubeTargetState targetState)
     this->setControlWord(targetState.getControlWord());
     ROS_INFO_THROTTLE(0.5, "\tWaiting for '%s': %s", targetState.getDescription().c_str(),
                       std::bitset<16>(this->getStatusWord()).to_string().c_str());
-    if (targetState.getState() == IMotionCubeTargetState::OPERATION_ENABLED.getState() && this->getState(this->getStatusWord()) == IMCState::fault)
+    if (targetState.getState() == IMotionCubeTargetState::OPERATION_ENABLED.getState()
+    && this->getState(this->getStatusWord()) == IMCState::fault)
     {
       ROS_FATAL("IMotionCube went to fault state while attempting to go to %s. Shutting down.",
                 targetState.getDescription().c_str());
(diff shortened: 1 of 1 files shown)


march_hardware/src/IMotionCube.cpp
```diff
@@ -467,6 +467,12 @@ bool IMotionCube::goToTargetState(IMotionCubeTargetState targetState)
     this->setControlWord(targetState.getControlWord());
     ROS_INFO_THROTTLE(0.5, "\tWaiting for '%s': %s", targetState.getDescription().c_str(),
                       std::bitset<16>(this->getStatusWord()).to_string().c_str());
+    if (targetState.getDescription() == "Operation Enabled" && this->getState(this->getStatusWord()) == IMCState::fault)
```
Ishadijcks: `targetState.getDescription() == "Operation Enabled"`
should be performed with an enum although I don't understand why it is needed at all?

Ishadijcks (dismissed):

march_hardware/src/IMotionCube.cpp
martijnvandermarel: I tried comparing with the enum but wasn't allowed to by the compiler for some reason. Will check later what error message was provided.

Reason why it is necessary at all is that the very first state transition in the goToOperationEnabled is from fault -> switch on disabled. Without the targetState = Operation Enabled check, an error would be thrown while it shouldn't be.

martijnvandermarel: What I tried: targetState == IMotionCubeTargetState::OPERATION_ENABLED
What compiler tells me: "error: no match for ‘operator==’ (operand types are ‘march4cpp::IMotionCubeTargetState’ and ‘const march4cpp::IMotionCubeTargetState’)"
Tried googling but couldn't really find a good fix, so then went with the next best thing which was comparing with the string.

Ishadijcks:
I think I have implemented the fix, can you try if it works?

martijnvandermarel:
Tested on exo, works. @Ishadijcks can you re-review?

TimBuckers (approved):

PR #164, opened 2019-10-14
Fix/pm 73 headers custom messages

RutgerVanBeek:
Closes PM-73

## Description

Custom messages in the March-iv repo were changed. An header was added. For the messages generated in the hardware-interface the time field of the header is filled. 
## Changes

At two points the time field of the header is filled.


Olavhaasie:
PM-73

Olavhaasie:
You should also add the header time stamp here

https://github.com/project-march/hardware-interface/blob/17757c1acc7e57fa1fe8302e05336f6fa357b053/march_hardware_state_controller/src/march_hardware_state_controller.cpp#L45-L52

RutgerVanBeek:
> You should also add the header time stamp here
> 
> https://github.com/project-march/hardware-interface/blob/17757c1acc7e57fa1fe8302e05336f6fa357b053/march_hardware_state_controller/src/march_hardware_state_controller.cpp#L45-L52

The message published here is the standard message. This is thus not a custom mesage.

Olavhaasie:
Okay we should also take a look at the usefulness of the `march_hardware_state_controller`

Olavhaasie (approved):

PR #169, opened 2019-10-28
Feature: add imu node

Olavhaasie:
Closes PM-64

## Description
Adds a new node `march_imu_manager_node` that connects to xsens MTws and publishes their data as `sensor_msgs::Imu` on `/march/imu/<id>`.

Currently this node is not launched during any launch sequence. We will start by testing it on the monitor PCs. The manager will wait for 1 MTw to connect and then starts measuring.

## Changes
* Add `march_imu_manager` package + `march_imu_manager_node` node
* Add some unit tests

**note**: you can skip all the files in `lib/` since those are library files copied from xsens.


Ishadijcks:
Why are you including a library in the source instead of a proper dependency?

Olavhaasie:
This dependency is not available through ROS packages and can only be downloaded from the xsens website. However this download link is missing from their website, so this is the easiest way to obtain the library :innocent:

Ishadijcks:
Release your own ROS package that wraps it :)

march_imu_manager/src/Mtw.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+const XsDataPacket* Mtw::getOldestPacket()
+{
+    std::unique_lock<std::mutex> lck(this->m_mutex);
+    XsDataPacket const * packet = &this->m_packetBuffer.front();
+    return packet;
+}
+
+void Mtw::deleteOldestPacket()
```
RutgerVanBeek, line 37: Wil je hier de packet niet ook gelijk returnen?

Olavhaasie: It is not possible to return the package and delete it, otherwise the package will be deleted from the queue and therefore there exists no valid reference to it. You could do this by `std::move` the packet.

march_imu_manager/src/WirelessMaster.cpp
```diff
@@ -0,0 +1,249 @@
...
+    {
+        if (mtw.second->dataAvailable())
+        {
+            const XsDataPacket* packet = mtw.second->getOldestPacket();
+
+            if (packet->containsCalibratedData())
+            {
+                sensor_msgs::Imu imu_msg;
+
+                imu_msg.header.frame_id = "imu_" + mtw.second->getId().toString().toStdString();
```
RutgerVanBeek: Hier moet niet de ID van de IMU komen, maar de ID van het frame waar de sensor bij hoort. Aangezien de IMU nog niet is toegevoegd aan de URDF heeft dit nog geen zin. Voor nu zou de frame_id dus leeggelaten moeten worden.

march_imu_manager/src/WirelessMaster.cpp
```diff
@@ -0,0 +1,249 @@
...
+void WirelessMaster::onConnectivityChanged(XsDevice* dev, XsConnectivityState newState)
+{
+    std::unique_lock<std::mutex> lck(this->m_mutex);
+    const uint32_t deviceId = dev->deviceId().toInt();
+    const std::string deviceIdString = dev->deviceId().toString().toStdString();
+    switch (newState)
+    {
+        case XCS_Disconnected:
+            ROS_INFO_STREAM("EVENT: MTW Disconnected -> " << deviceIdString);
+            this->m_connectedMtws.erase(deviceId);
```
RutgerVanBeek, line 189: Wil je hier niet een warning toevoegen?

march_imu_manager/src/WirelessMaster.cpp
Olavhaasie: I'll remove it for now

RutgerVanBeek (changes requested):
When the two comments are implemented, I approve. I did not check the installation of the library, but the IMU's worked as expected.

RutgerVanBeek (approved):

PR #170, opened 2019-10-30
new endstop values lhaa lkfe

RPKrijnen:
Closes PM-

## Description
Recalibration for lkfe and lhaa. After the left leg has been taken apart, recalibration was needed to fix a discrepancy in the lhaa home stand position. During the fix another discrepancy was found in the lkfe. This pull request adjusts the min, max and zero positions of these joints in the march4.yaml file.

## Changes
adjusted the lhaa and lhfe min, max and zero positions in the march4.yaml

### DIFF
--- a/march_hardware_builder/robots/march4.yaml
+++ b/march_hardware_builder/robots/march4.yaml
@@ -18,7 +18,7 @@ march4:
         temperatureges: {slaveIndex: 7, byteOffset: 0}
         imotioncube:
           slaveIndex: 8
-          encoder: {resolution: 12, minPositionIU: 1010, maxPositionIU: 1401, zeroPositionIU: 1211, safetyMarginRad: 0.05}
+          encoder: {resolution: 12, minPositionIU: 1035, maxPositionIU: 1438, zeroPositionIU: 1242, safetyMarginRad: 0.05}
 
     - right_hip_fe:
         actuationMode: position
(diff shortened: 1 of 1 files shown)


Roelemans (approved):
Since it works on the exo it should be fine :+1:. Was the right hip AA not giving any problems?

Olavhaasie (commented):
Is there somewhere documented on how to calibrate the joints? And are there any more places where this should be changed?

RPKrijnen:
> Since it works on the exo it should be fine +1. Was the right hip AA not giving any problems?

We thought it was at first, but that was due to using the wrong workspace. After switching to catkin_ws with the new yaml everything seemed to work just like the trainings with march 4 i.e. occasional ethercat errors, but nothing substantial.

RPKrijnen:
> Is there somewhere documented on how to calibrate the joints? And are there any more places where this should be changed?

This confluence page gives some insight into the matter: https://confluence.projectmarch.nl:8443/display/41TECH/Setup+Test+Joint+M3
It however describes that an approximation is good enough. Yesterday I verified with the old BaCo department that it would be better to instead, move to the endstops, record the value and subtract from it a set angle (which has to be converted to internal units). This would be the same method that M3 used.

Roelemans:
Is it a good idea to create a confluence page dedicated to this topic? It seems rather important.

Ishadijcks:
The endstops you can properly record, to calibrate 0 you can either move it to 0 degrees physically and read the IU value. Or calculate it based on the ROM of the joint and the angle at which it is fixed.

These should give the same answer ofcourse but they don't. Use whichever you are more comfortable with.

PR #171, opened 2019-10-31
Change levels of logs

Olavhaasie:
And make messages more consistent with prepending the joint names within
brackets.

Closes PM-95

## Description
Downgraded some log levels from **info** to **debug** to reduce spam and make messages more consistent on joint names. The hardware interface did not require any kind of action or service ordering, since this is managed by the `controller_manager`. This way we can check if the hardware interface has launched by calling the `ListControllers` service.

## Changes
* Downgraded some `ROS_INFO` to `ROS_DEBUG`

### DIFF
--- a/march_hardware/src/EtherCAT/EthercatMaster.cpp
+++ b/march_hardware/src/EtherCAT/EthercatMaster.cpp
@@ -183,7 +183,7 @@ int EthercatMaster::receiveProcessData()
   int wkc =  ec_receive_processdata(EC_TIMEOUTRET);
   if (wkc < this->expectedWKC)
   {
-    ROS_WARN("Working counter lower than expected. EtherCAT connection may not be optimal");
+    ROS_WARN_THROTTLE(1, "Working counter lower than expected. EtherCAT connection may not be optimal");
   }
   return wkc;
 }
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -465,7 +465,7 @@ bool IMotionCube::goToTargetState(IMotionCubeTargetState targetState)
   while (!targetState.isReached(this->getStatusWord()))
   {
     this->setControlWord(targetState.getControlWord());
-    ROS_DEBUG_THROTTLE(0.5, "\tWaiting for '%s': %s", targetState.getDescription().c_str(),
+    ROS_INFO_DELAYED_THROTTLE(2, "\tWaiting for '%s': %s", targetState.getDescription().c_str(),
                       std::bitset<16>(this->getStatusWord()).to_string().c_str());
     if (targetState.getState() == IMotionCubeTargetState::OPERATION_ENABLED.getState() &&
         this->getState(this->getStatusWord()) == IMCState::fault)
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -461,23 +461,23 @@ std::string IMotionCube::parseDetailedError(uint16 detailedError)
 
 bool IMotionCube::goToTargetState(IMotionCubeTargetState targetState)
 {
-  ROS_INFO("\tTry to go to '%s'", targetState.getDescription().c_str());
+  ROS_DEBUG("\tTry to go to '%s'", targetState.getDescription().c_str());
   while (!targetState.isReached(this->getStatusWord()))
   {
     this->setControlWord(targetState.getControlWord());
-    ROS_INFO_THROTTLE(0.5, "\tWaiting for '%s': %s", targetState.getDescription().c_str(),
+    ROS_DEBUG_THROTTLE(0.5, "\tWaiting for '%s': %s", targetState.getDescription().c_str(),
                       std::bitset<16>(this->getStatusWord()).to_string().c_str());
     if (targetState.getState() == IMotionCubeTargetState::OPERATION_ENABLED.getState() &&
         this->getState(this->getStatusWord()) == IMCState::fault)
     {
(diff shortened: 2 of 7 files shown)


Ishadijcks:
> Downgrade some ROS_INFO to ROS_DEBUG

I agree that these should be debugs, are you wouldn't need them if everything is working correctly. However as you saw yesterday, it doesn't :)

I think you'll need this information more than you think, but we will see :)

Olavhaasie:
> > Downgrade some ROS_INFO to ROS_DEBUG
> 
> I agree that these should be debugs, are you wouldn't need them if everything is working correctly. However as you saw yesterday, it doesn't :)
> 
> I think you'll need this information more than you think, but we will see :)

I agree that we need this information. However this is debug information, so if there is something wrong we will inspect the debug log level. Debug logs are also stored in `~/.ros/log`, so we can inspect them after the launch.

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -183,7 +183,7 @@ int EthercatMaster::receiveProcessData()
   int wkc =  ec_receive_processdata(EC_TIMEOUTRET);
   if (wkc < this->expectedWKC)
   {
-    ROS_WARN("Working counter lower than expected. EtherCAT connection may not be optimal");
+    ROS_WARN_THROTTLE(1, "Working counter lower than expected. EtherCAT connection may not be optimal");
```
Roelemans, line 186: Good fix!

march_hardware/src/IMotionCube.cpp
```diff
@@ -461,23 +461,23 @@ std::string IMotionCube::parseDetailedError(uint16 detailedError)
 
 bool IMotionCube::goToTargetState(IMotionCubeTargetState targetState)
 {
-  ROS_INFO("\tTry to go to '%s'", targetState.getDescription().c_str());
+  ROS_DEBUG("\tTry to go to '%s'", targetState.getDescription().c_str());
   while (!targetState.isReached(this->getStatusWord()))
   {
     this->setControlWord(targetState.getControlWord());
-    ROS_INFO_THROTTLE(0.5, "\tWaiting for '%s': %s", targetState.getDescription().c_str(),
+    ROS_DEBUG_THROTTLE(0.5, "\tWaiting for '%s': %s", targetState.getDescription().c_str(),
```
Roelemans: Without this you have no indication that anything is happening if the exo gets stuck in this loop. Either keep it in info, or throw info messages/warnings/an error after ~2-10 seconds trying.

march_hardware/src/Joint.cpp
```diff
@@ -25,13 +25,13 @@ void Joint::prepareActuation()
                  this->name.c_str(), this->getActuationMode().toString().c_str());
   if (this->allowActuation)
   {
-    ROS_INFO("Preparing joint %s for actuation", this->name.c_str());
+    ROS_INFO("[%s] Preparing for actuation", this->name.c_str());
```
Roelemans, line 28: Is this not a strange format? Now the message will look like 
"[LHAA] Preparing for actuation"

march_hardware/src/Joint.cpp
```diff
@@ -44,9 +44,9 @@ void Joint::resetIMotionCube()
 
 void Joint::actuateRad(float targetPositionRad)
 {
-  ROS_ASSERT_MSG(this->allowActuation, "Joint %s is not allowed to actuate, "
-                                       "yet its actuate method has been "
-                                       "called.",
+  ROS_ASSERT_MSG(this->allowActuation,
+                 "Joint %s is not allowed to actuate, "
```
Roelemans, line 48: This does not follow the format for joint from the other messages you changed.

march_hardware_interface/launch/testjoint_linear.launch
```diff
@@ -10,7 +10,7 @@
             name="record"
             type="record"
             output="screen"
-            args="-o /home/projectmarch/Documents/logs/Testjoint.bag /march/controller/trajectory/state /clock"
+            args="-q -o /home/projectmarch/Documents/logs/Testjoint.bag /march/controller/trajectory/state /clock"
```
Roelemans, line 13: NICE!

Roelemans (changes requested):

march_hardware/src/Joint.cpp
Olavhaasie: it will look like:

    [left_hip_aa] Preparing for actuation

march_hardware/src/Joint.cpp
Olavhaasie: True, that is because this assert message is being tested and for some weird reason adding brackets (`[%s]`) fails the test... :upside_down_face:

march_hardware/src/IMotionCube.cpp
Olavhaasie: I changed to a `ROS_INFO_DELAYED_THROTTLE(2,..)`, which will print every 2 seconds starting at 2 seconds.

march_hardware/src/Joint.cpp
Roelemans: But aren't those brackets and the captial P strange?

march_hardware/src/Joint.cpp
Roelemans: :joy:

march_hardware/src/Joint.cpp
Olavhaasie: @RutgerVanBeek hjelp

RutgerVanBeek: I don't care

RutgerVanBeek (approved):

Roelemans (approved):

PR #172, opened 2019-11-05
synchronzing joint orders

RutgerVanBeek:
Closes PM-72

## Description

A lot of messages contain information per joint. The order in which this information is presented is different per message. This is inconvenient when you are looking into the log files or when rewriting the log files to a different format. The messages connected to the controller have an alphabetical order as well now. This is also the order of the joints when they are taken from the URDF. 
## Changes

The joint order in the controller yaml is changed.


Olavhaasie (approved):
Tested on exo and it works :1st_place_medal:

PR #174, opened 2019-11-11
Fix hardware launch file

Olavhaasie:
Partly closes PM-96

## Changes

* Rename to hardware.launch
* Remove march3_hardware.launch
* Give robot option
* Remove `output="screen"`


march_hardware_interface/launch/hardware.launch
```diff
@@ -0,0 +1,28 @@
+<launch>
+    <arg name="config" default="march4" doc="The robot to run. Can be: march3, march4, test_joint_linear, test_joint_rotational."/>
+
+    <rosparam file="$(find march_hardware_interface)/config/$(arg config)/controllers.yaml" command="load"/>
+
+    <param name="robot_description"
+           command="$(find xacro)/xacro --inorder '$(find march_description)/urdf/march-iv.xacro'" />
```
Roelemans: You always use the march-iv.xacro, even when launching march3. I believe this will result in the march3 thinking it has 8 joints instead of 6.

Roelemans (changes requested):

march_hardware_interface/launch/hardware.launch
```diff
@@ -0,0 +1,28 @@
+<launch>
+    <arg name="config" default="march4" doc="The robot to run. Can be: march3, march4, test_joint_linear, test_joint_rotational."/>
```
Ishadijcks: Config is a non-descriptive name

march_hardware_interface/launch/hardware.launch
Ishadijcks: Should just use the variable defined above

Roelemans (approved):

PR #176, opened 2019-11-13
Revert ethercat_grant and add setcap

Olavhaasie:
## Description
This PR reverts the `ethercat_grant` launch prefix, since the executable was not able to dynamically link the libraries. See https://github.com/shadow-robot/ethercat_grant/issues/4. Therefore I reverted this change and added an `if()` statement that will only grant the raw ethernet ports permission if the environment variable `ETHERCAT_GRANT` is set to `1`. So the exoskeleton will have 

    export ETHERCAT_GRANT=1

added to its `.bashrc`. Furthermore the exoskeleton will have sudo privileges for the `setcap` utitlity, so that it does not require a password. You probably don't want to set this variable on your development machine, otherwise you have to type in your password :computer: 

## Changes
* remove `launch-prefix="ethercat_grant"`
* Added `setcap` command post build when `ETHERCAT_GRANT` is set to `1`

### DIFF
--- a/march_hardware/CMakeLists.txt
+++ b/march_hardware/CMakeLists.txt
@@ -85,8 +85,19 @@ install(TARGETS slave_count_check
     RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 )
 
+if($ENV{ETHERCAT_GRANT})
+    message("Granting raw ethernet port")
+    # Give sudo access to ethernet port
+    add_custom_command(
+        TARGET slave_count_check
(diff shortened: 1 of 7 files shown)


Ishadijcks:
You can also remove password requirements for certain sudo commands on your local machine. Don't forget to add these ethercat grant to other machines that use it, such as the test joint and the backup master.

RutgerVanBeek (commented):
LGTM, should be tested on the exo for merging

Roelemans:
Good luck

PR #177, opened 2019-11-13
Generalize IMU topics

RutgerVanBeek:
Necessary for  PM-103

## Description

Instead of creating a topic for every IMU id it is nice to have just one topic devices can than be distinguished based on their frame_id in the header. I have set this to "imu-link" from the URDF as a standard now. If we have different places for different IMU's, then they should have different frame_id's. 
## Changes

Set the frame_id standard to "imu_link"
Publish on /march/imu


march_imu_manager/src/WirelessMaster.cpp
```diff
@@ -205,7 +206,7 @@ void WirelessMaster::onConnectivityChanged(XsDevice* dev, XsConnectivityState ne
                 this->m_connectedMtws.insert(std::make_pair(deviceId, std::unique_ptr<Mtw>(new Mtw(dev))));
 
                 ros::Publisher publisher =
-                    this->m_node->advertise<sensor_msgs::Imu>("march/imu/" + deviceIdString, 10);
+                    this->m_node->advertise<sensor_msgs::Imu>("march/imu/", 10);
```
Olavhaasie:
```suggestion
                    this->m_node->advertise<sensor_msgs::Imu>("/march/imu/", 10);
```

Olavhaasie (changes requested):

Olavhaasie (approved):
:+1:

PR #182, opened 2019-12-06
Fix/test setup

Olavhaasie:
## Description
The controller names were incorrect and could therefore not be actuated by rqt trajectory plugin. And the I made the test joint names consistent with the rest.

## Changes
* Removed `/march/` namespace from controller names
* Renamed `testjoint_linear` and `testjoint_rotational` to `test_joint_linear` and `test_joint_rotational`

### DIFF
--- a/march_hardware_builder/include/march_hardware_builder/AllowedRobot.h
+++ b/march_hardware_builder/include/march_hardware_builder/AllowedRobot.h
@@ -11,8 +11,8 @@ public:
   {
     march4,
     march3,
-    testjoint_rotational,
-    testjoint_linear,
+    test_joint_rotational,
+    test_joint_linear,
     pdb,
   };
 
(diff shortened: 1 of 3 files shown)


bjornminderman (commented):
Why did you remove the /march/ workspace from the controller names?

Olavhaasie:
> Why did you remove the /march/ workspace from the controller names?

The controller parameters are uploaded to the parameter server on `/march/controller/` and that's why it said `/march/controller/trajectory`, except the `controller_spawner` node was already launched in the `/march` namespace. So the controller name would end up being `/march//march/controller/trajectory`. We discovered this when we wanted to actuate the test joint using the rqt joint trajectory controller plugin. I have tested this on the exo, and now the controllers are correctly loaded and named `/march/controller/trajectory`

bjornminderman (approved):

Roelemans (approved):

PR #183, opened 2019-12-08
Fix/log fatal

Olavhaasie:
## Description
During the training I found that most IMotionCube errors were not logged using `ROS_FATAL` when they threw an exception. Whether they should throw an exception is for another time.

## Changes
* Add `ROS_FATAL` to `iMotionCubeStateCheck` and `outsideLimitsCheck`
* Add `ROS_INFO` when all joints have been actuated


RutgerVanBeek (approved):

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -217,6 +217,7 @@ void MarchHardwareInterface::init()
       joint.prepareActuation();
     }
   }
+  ROS_INFO("Successfully actuated all joints");
```
Roelemans, line 220: Does this mean exo is ready for action?

Roelemans (approved):

PR #184, opened 2019-12-08
Refactor hardware builder

Olavhaasie:
## Description
Refactored the hardware builder to be more consistent. And use ros unit tests instead of rostest.

## Changes
* Remove `rostest` setup, since the `march_hardware_builder` is not a node
* Rename files to under_score
* Make the `march_hardware_interface` node return error code when no robot argument is given
* Remove too many log statements in `createPowerDistributionBoard`
* Change some info logs to debug logs, since those always happen
* Make code adhere to the ROS cpp style guide


RutgerVanBeek (approved):
This still should be tested Tuesday afternoon in the software test.

Roelemans (approved):

PR #186, opened 2019-12-12
Fix ethercat_grant

Olavhaasie:
## Description
Fixes the problem that the executables were not able to find the shared libraries through `ethercat_grant`. Should be tested on exo first 

## Changes
* Added RPATH generation on `march_hardware_interface` executable


RutgerVanBeek (approved):
I think we should test this on the exo before merging.

Roelemans (approved):
I think we should test this on the exo before merging.

Joris-JW (approved):
i think we should test this on exo before merging.

PR #188, opened 2019-12-13
Fix/pm 196 fix return value initation imc check

Joris-JW:
Closes PM-196

## Description
When initiating the IMC with different SDO-messages the return value is the workingcounter of the SDO. Currently the program does a bitwise-and with a boolean which states the succes variable. The working counter could be an even number which results in the ROS_ASSERT_MSG being initiated. This is changed to checking if the return value is 0, this is the error statement and the program will raise an error. 

## Changes
* Removed the succes boolean from the initiation of the IMC
* Added a check if the working counter is 0 in the SDO write functions


Roelemans (dismissed):
Looks good! Has this been tested? Does it improve things?

RutgerVanBeek (dismissed):
Nice work. Of course test on exo first

march_hardware/src/EtherCAT/EthercatSDO.cpp
```diff
@@ -14,23 +14,39 @@
 
 namespace march4cpp
 {
+int ReceivedWorkingCounter;
```
Olavhaasie: You don't have to use a global variable here, since you only use it local in the methods

Olavhaasie (changes requested):

Olavhaasie (approved):

Olavhaasie:
Closed since this fix returns errors from the sdo writes. Will be looked at in PM-229

PR #191, opened 2019-12-17
Feature/pm 203 new imc initialization

Joris-JW:
Closes PM-203

## Description
This branch removed the resetIMCuntilTheyWork function to be replaced by a single reset of all th imotion cubes. when initializing the encoders and IMC's it is not uncommon to read an angle of 0 because the joints have not been actualized. After the reset all the joints are actualized (the ticks) when the ethercat master is restarted. All the encoders should have a valid value.

## Changes
* Removed the reset IMC until they work
* Added a single reset function
* Added an extra try catch to make sure ethercat has been deconstructed properly

## Tested
* Tested on the rotational test joint
* Tested on exo


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -317,33 +317,34 @@ void MarchHardwareInterface::write(const ros::Duration& elapsed_time)
...
-      // TODO(Martijn) check if you need to reset all joints.
-      for (int i = 0; i < num_joints_; ++i)
-      {
-        march4cpp::Joint joint = marchRobot.getJoint(joint_names_[i]);
-        joint.resetIMotionCube();
-      }
-      ROS_INFO("Restarting EtherCAT");
-      marchRobot.stopEtherCAT();
-      marchRobot.startEtherCAT();
+      ROS_WARN_THROTTLE(1, "[%s] still has encoder value 0 after initiation", joint_names_[i].c_str());
```
RutgerVanBeek: Shouldn't this be a fatal? Can we safely start actuating when there is a wrong encoder value?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -317,33 +317,34 @@ void MarchHardwareInterface::write(const ros::Duration& elapsed_time)
...
-  while (!encoderSetCorrectly)
+  for (int i = 0; i < num_joints_; ++i)
   {
-    encoderSetCorrectly = true;
-    for (int i = 0; i < num_joints_; ++i)
+    march4cpp::Joint joint = marchRobot.getJoint(joint_names_[i]);
+    if (joint.getAngleIU() == 0)
     {
-      march4cpp::Joint joint = marchRobot.getJoint(joint_names_[i]);
-      if (joint.getAngleIU() == 0)
```
RutgerVanBeek, line 330: Personally I would increase the range because March 4 has experience with encoder resets of -2, -1, 0, 1, 2

march_hardware_interface/src/march_hardware_interface_node.cpp
```diff
@@ -32,10 +32,19 @@ int main(int argc, char** argv)
...
-    march.update(rate.expectedCycleTime());
-    rate.sleep();
+    try
+    {
+      march.update(rate.expectedCycleTime());
+      rate.sleep();
+    }
+    catch (const std::exception& e)
+    {
+      ROS_FATAL("Hardware interface caught an exception during init: %s", e.what());
```
Olavhaasie: :upside_down_face: 
```suggestion
      ROS_FATAL("Hardware interface caught an exception during update: %s", e.what());
```

march_hardware_interface/src/march_hardware_interface.cpp
Olavhaasie: There is also an extra check for this in `IMotionCube::goToOperationEnabled`

Olavhaasie (changes requested):

Roelemans (dismissed):
Has this been tested on exo?

RutgerVanBeek (dismissed):

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -317,33 +317,35 @@ void MarchHardwareInterface::write(const ros::Duration& elapsed_time)
...
-      {
-        ROS_ERROR("[%s] Failed (encoder reset)", joint_names_[i].c_str());
-        encoderSetCorrectly = false;
-      }
+      ROS_WARN_THROTTLE(1, "[%s] has encoder value 0 ", joint_names_[i].c_str());
```
Olavhaasie: I don't think we want to throttle this, since it should be printed for every joint that has this issue. I think this whole loop happens within 1 second, so it will always print 0 or 1 joint.
```suggestion
      ROS_WARN("[%s] has encoder value 0 ", joint_names_[i].c_str());
```

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -317,33 +317,35 @@ void MarchHardwareInterface::write(const ros::Duration& elapsed_time)
...
-      // TODO(Martijn) check if you need to reset all joints.
-      for (int i = 0; i < num_joints_; ++i)
-      {
-        march4cpp::Joint joint = marchRobot.getJoint(joint_names_[i]);
-        joint.resetIMotionCube();
-      }
-      ROS_INFO("Restarting EtherCAT");
-      marchRobot.stopEtherCAT();
-      marchRobot.startEtherCAT();
+      ROS_WARN_THROTTLE(1, "Joint: [%s] has angle-value of: %i. Which is within boundary of lower: %i and upper: %i",
```
Olavhaasie: Same here
```suggestion
      ROS_WARN("Joint: [%s] has angle-value of: %i. Which is within boundary of lower: %i and upper: %i",
```

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -317,33 +317,35 @@ void MarchHardwareInterface::write(const ros::Duration& elapsed_time)
...
+void MarchHardwareInterface::initiateIMC()
 {
-  bool encoderSetCorrectly = false;
-
-  while (!encoderSetCorrectly)
+  for (int i = 0; i < num_joints_; ++i)
   {
-    encoderSetCorrectly = true;
-    for (int i = 0; i < num_joints_; ++i)
+    march4cpp::Joint joint = marchRobot.getJoint(joint_names_[i]);
```
Olavhaasie: It is probably nicer to loop using a foreach like so:
```c++
for (march4cpp::Joint& joint : marchRobot.jointList) {
...
}
```
And then even it would be nicer to make the `jointList` member private and create a `getJoints` method. And also add `using march4cpp::Joint;` at the top of the file, so that hideous namespace does not have to be defined every time. :slightly_smiling_face:

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -317,33 +317,35 @@ void MarchHardwareInterface::write(const ros::Duration& elapsed_time)
...
-      }
+      ROS_WARN_THROTTLE(1, "[%s] has encoder value 0 ", joint_names_[i].c_str());
     }
-    if (!encoderSetCorrectly)
+  }
+
+  ROS_INFO("Resetting all IMC on initialization");
+  for (int i = 0; i < num_joints_; ++i)
+  {
+    march4cpp::Joint joint = marchRobot.getJoint(joint_names_[i]);
```
Olavhaasie: Again a foreach loop here

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -317,33 +317,35 @@ void MarchHardwareInterface::write(const ros::Duration& elapsed_time)
...
+    joint.resetIMotionCube();
+  }
+
+  ROS_INFO("Restarting EtherCAT");
+  marchRobot.stopEtherCAT();
+  marchRobot.startEtherCAT();
+
+  for (int i = 0; i < num_joints_; ++i)
+  {
+    march4cpp::Joint joint = marchRobot.getJoint(joint_names_[i]);
```
Olavhaasie: And again a foreach loop

Olavhaasie (changes requested):

Olavhaasie (approved):

RutgerVanBeek (approved):

PR #192, opened 2020-01-06
Fix/pm 229/rewrite pdo mapping

Joris-JW:
Closes PM-229

## Description
This PR has changed the PDO mapping sequence to match the example given in the IMC-manual at page 51. In general first the PDO-message is deactivated to prepare the changes. Next the objects that fit into the 64-bit message will be written in a specific format described by the manual. When these steps are completed successful the sync manager of the IMC is deactivated and changed accordingly. Both the PDO and sync manager are activated after the changes have been written.
  
Also added branch "fix/PM-196-fix-return-value-initiation-imc-check for testing" because this test originally found the problem of a wrong PDO-sequence. 

## Changes
* Changed PDO mapping sequence
* Updated code and added comments
* Merged fix/PM-196-fix-return-value-initiation-imc-check for testing

## Tested and verified  
* Tested on the rotational test joint.
* Tested on exo!!


march_hardware/src/IMotionCube.cpp
```diff
@@ -76,41 +76,39 @@ void IMotionCube::validateMosiPDOs()
...
 
   // Quick stop deceleration
-  success &= sdo_bit32(slaveIndex, 0x6085, 0, 0x7FFFFFFF);
+  sdo_bit32(slaveIndex, 0x6085, 0, 0x7FFFFFFF);
 
   // set the ethercat rate of encoder in form x*10^y
-  success &= sdo_bit8(slaveIndex, 0x60C2, 1, ecatCycleTime);
-  success &= sdo_bit8(slaveIndex, 0x60C2, 2, -3);
-
-  ROS_ASSERT_MSG(success, "Writing initial settings to IMC %d failed", this->slaveIndex);
```
Ciliaclaij, line 113: So we don't check anymore if writing initial settings succeeds. We encountered this error a few times earlier, so maybe we still need this check. Is there a reason you removed it?

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -57,24 +59,29 @@ enum class IMCObjectName
...
+
   std::map<IMCObjectName, IMCObject> PDOObjects;
   std::map<IMCObjectName, IMCObject> allObjects;
   std::vector<std::pair<IMCObjectName, IMCObject>> sortedPDOObjects;
   std::map<IMCObjectName, int> byteOffsets;
 
-  const int bitsPerReg = 64;
-  const int nrofRegs = 4;
-  const int objectSizes[3] = { 8, 16, 32 };
+  const int bitsPerReg = 64;                 // Maximum amount of bits that  constructed in one PDO message.
```
Olavhaasie:
```suggestion
  const int bitsPerReg = 64;                 // Maximum amount of bits that can be constructed in one PDO message.
```

march_hardware/src/EtherCAT/EthercatSDO.cpp
```diff
@@ -18,19 +18,36 @@ namespace march4cpp
 int sdo_bit8(int slave, uint32_t index, uint8_t sub, uint8_t value)
 {
   ROS_DEBUG("sdo_bit8: slaveIndex %i, reg 0x%X, subindex %i, value 0x%X", slave, index, sub, value);
-  return ec_SDOwrite(slave, index, sub, FALSE, 1, &value, EC_TIMEOUTRXM);
+  int ReceivedWorkingCounter = ec_SDOwrite(slave, index, sub, FALSE, 1, &value, EC_TIMEOUTRXM);
```
Olavhaasie: Name your local variables **under_scored** (https://wiki.ros.org/CppStyleGuide#Variables)
```suggestion
  int received_working_counter = ec_SDOwrite(slave, index, sub, FALSE, 1, &value, EC_TIMEOUTRXM);
```

march_hardware/src/IMotionCube.cpp
Olavhaasie: It is still being checked, however, it happens inside the `sdo_bit` functions

march_hardware/src/PDOmap.cpp
```diff
@@ -8,124 +8,143 @@ namespace march4cpp
...
+  this->allObjects[IMCObjectName::MotorPosition] = IMCObject(0x2088, 32);
+  this->allObjects[IMCObjectName::ControlWord] = IMCObject(0x6040, 16);
+  this->allObjects[IMCObjectName::TargetPosition] = IMCObject(0x607A, 32);
+  this->allObjects[IMCObjectName::TargetTorque] = IMCObject(0x6071, 16);
+  this->allObjects[IMCObjectName::QuickStopDeceleration] = IMCObject(0x6085, 32);
+  this->allObjects[IMCObjectName::QuickStopOption] = IMCObject(0x605A, 16);
 }
 
-void PDOmap::addObject(IMCObjectName objectname)
+void PDOmap::addObject(IMCObjectName objectName)
```
Olavhaasie: Actually parameters should be **under_score** (https://wiki.ros.org/CppStyleGuide#Function_.2BAC8_Methods)

march_hardware/src/PDOmap.cpp
```diff
@@ -8,124 +8,143 @@ namespace march4cpp
...
 {
   this->sortPDOObjects();
-  int reg;
-  int SMAddress;
+
   if (direction == dataDirection::miso)
   {
-    reg = 0x1A00;
-    SMAddress = 0x1C13;
+    ROS_INFO("Mapping miso");
```
Olavhaasie: This info is a bit useless as it gives no detailed information, and if it does it should be `ROS_DEBUG`

Olavhaasie (changes requested):

march_hardware/src/EtherCAT/EthercatSDO.cpp
```diff
@@ -18,19 +18,36 @@ namespace march4cpp
...
+              value, slave);
+  }
+  return ReceivedWorkingCounter;
 }
 
 int sdo_bit16(int slave, uint32_t index, uint8_t sub, uint16_t value)
 {
   ROS_DEBUG("sdo_bit16: slaveIndex %i, reg 0x%X, subindex %i, value 0x%X", slave, index, sub, value);
-  return ec_SDOwrite(slave, index, sub, FALSE, 2, &value, EC_TIMEOUTRXM);
+  int ReceivedWorkingCounter = ec_SDOwrite(slave, index, sub, FALSE, 2, &value, EC_TIMEOUTRXM);
```
RPKrijnen: Working counters are not by definition 0 when nothing has been read/written to the register. The strategy should instead be; read an expected working counter and compare the counter you receive with the one you expect. Only if the counter has not been incremented it should fail. See the following links for more info; https://infosys.beckhoff.com/english.php?content=../content/1033/tc3_io_intro/1257993099.html&id=3196541253205318339
https://github.com/OpenEtherCATsociety/SOEM/blob/master/test/intime/ec_master/ec_master.c

march_hardware/src/IMotionCube.cpp
```diff
@@ -76,41 +76,39 @@ void IMotionCube::validateMosiPDOs()
 void IMotionCube::writeInitialSettings(uint8 ecatCycleTime)
 {
   ROS_DEBUG("IMotionCube::writeInitialSettings");
-  bool success = true;
-  // sdo_bit32(slaveIndex, address, subindex, value);
 
   if (this->actuationMode == ActuationMode::unknown)
   {
     throw std::runtime_error("Cannot write initial settings to IMotionCube as it has actuation mode of unknown");
   }
```
RPKrijnen, line 84: I like to see the addition of the register 6007. This enables the iMC with functionality to react to an etherCAT communication loss. 
```suggestion
sdo_bit8(slaveIndex, 0x6007, 0, 1);
```

RPKrijnen (changes requested):
Good to see you have used the iMC manual as a guide! I would like to see the addition of the register 6007h in the IMotionCube.cpp file and a different strategy of using the working counter. see the following for some inspiration: https://github.com/OpenEtherCATsociety/SOEM/blob/master/test/intime/ec_master/ec_master.c

march_hardware/src/EtherCAT/EthercatSDO.cpp
Joris-JW: I absolutely agree with u to not only check the working on a possible zero value but also thresholds. However my suggestion would be to add these changes in branch:  "fix/PM-196-fix-return-value-initiation-imc-check for testing" . The current check (for 0) is only to check wether an error occurs in writing or reading. If the PDO map is compatible with this version we may add more checks to have a better level of certainty.

RPKrijnen (dismissed):

march_hardware/src/PDOmap.cpp
Joris-JW: changed all the class variables in the PDO-map files to match the cpp style guide

march_hardware/src/PDOmap.cpp
Joris-JW: removed, redundant statements

march_hardware/src/IMotionCube.cpp
Joris-JW: I will add and test this in another branch. This branch has been tested with exo and should be merged as quickly as possible.

march_hardware/src/IMotionCube.cpp
Olavhaasie, line 113: With this assert removed the `sdo_bit` functions will now print a fatal but not actually error or anything and just continue. You should still add an exception throw. It will be nicer handled in the future, when we discuss this in PM-52

march_hardware/src/PDOmap.cpp
```diff
@@ -8,125 +8,142 @@ namespace march4cpp
...
-  if (this->allObjects.count(objectname) != 1)
+  if (this->all_objects.count(object_name) != 1)
   {
     ROS_WARN("IMC object does not exist (yet), or multiple exist");
     return;
   }
-  else if (this->PDOObjects.count(objectname) != 0)
+  else if (this->PDO_objects.count(object_name) != 0)
   {
     ROS_WARN("IMC object is already added to PDO map");
```
Olavhaasie: I think it is useful to add the `object_name` variable to the warning.

march_hardware/src/PDOmap.cpp
```diff
@@ -8,125 +8,142 @@ namespace march4cpp
...
   {
-    reg = 0x1A00;
-    SMAddress = 0x1C13;
+    return configurePDO(slaveIndex, 0x1A00, 0x1C13);
   }
   else if (direction == dataDirection::mosi)
   {
-    reg = 0x1600;
-    SMAddress = 0x1C12;
+    return configurePDO(slaveIndex, 0x1600, 0x1C12);
```
Olavhaasie: These are magic numbers, you should add them as constants to `PDOmap`

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -57,24 +59,29 @@ enum class IMCObjectName
 class PDOmap
 {
 public:
-  // Constructor
+  /** Initiate all the entered IMC objects to prepare the PDO.*/
   PDOmap();
-
-  void addObject(IMCObjectName objectname);
+  void addObject(IMCObjectName object_name);
   std::map<IMCObjectName, int> map(int slaveIndex, dataDirection direction);
```
Olavhaasie: You changed the parameter name of `addObject`, you should also make these **under_score**

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -57,24 +59,29 @@ enum class IMCObjectName
...
 
 private:
-  void initAllObjects();
+  /** This function is used to sort the objects in the all_objects according to data length.*/
   void sortPDOObjects();
+
+  /** This function configures the PDO in the IMC using the given base register address and sync manager address.*/
+  std::map<IMCObjectName, int> configurePDO(int slaveIndex, int baseRegister, int baseSyncManager);
```
Olavhaasie: Same for these parameter names

march_hardware/src/PDOmap.cpp
```diff
@@ -8,125 +8,142 @@ namespace march4cpp
...
+  this->all_objects[IMCObjectName::DCLinkVoltage] = IMCObject(0x2055, 16);
+  this->all_objects[IMCObjectName::DriveTemperature] = IMCObject(0x2058, 16);
+  this->all_objects[IMCObjectName::ActualTorque] = IMCObject(0x6077, 16);
+  this->all_objects[IMCObjectName::CurrentLimit] = IMCObject(0x207F, 16);
+  this->all_objects[IMCObjectName::MotorPosition] = IMCObject(0x2088, 32);
+  this->all_objects[IMCObjectName::ControlWord] = IMCObject(0x6040, 16);
+  this->all_objects[IMCObjectName::TargetPosition] = IMCObject(0x607A, 32);
+  this->all_objects[IMCObjectName::TargetTorque] = IMCObject(0x6071, 16);
+  this->all_objects[IMCObjectName::QuickStopDeceleration] = IMCObject(0x6085, 32);
+  this->all_objects[IMCObjectName::QuickStopOption] = IMCObject(0x605A, 16);
```
Olavhaasie: `all_objects` is the same for every instance of `PDOmap` so you could make this a `static` variable to make it shared between different instances. Then it will construct once and not every time a new `PDOmap` is made

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -57,24 +59,29 @@ enum class IMCObjectName
...
-  std::map<IMCObjectName, IMCObject> PDOObjects;
-  std::map<IMCObjectName, IMCObject> allObjects;
-  std::vector<std::pair<IMCObjectName, IMCObject>> sortedPDOObjects;
-  std::map<IMCObjectName, int> byteOffsets;
 
-  const int bitsPerReg = 64;
-  const int nrofRegs = 4;
-  const int objectSizes[3] = { 8, 16, 32 };
+  std::map<IMCObjectName, IMCObject> PDO_objects;
+  std::map<IMCObjectName, IMCObject> all_objects;
```
Olavhaasie: The default [`map`](http://www.cplusplus.com/reference/map/map/) type is a weakly ordered map. However, you use no ordering in the `PDO_objects` and `all_objects` member variable. Therefore, you should make this an `std::unordered_map`, which improves performance.

march_hardware/src/PDOmap.cpp
```diff
@@ -8,125 +8,142 @@ namespace march4cpp
...
     return;
   }
   else
   {
-    this->PDOObjects[objectname] = this->allObjects[objectname];
+    this->PDO_objects[object_name] = this->all_objects[object_name];
   }
 }
 
 std::map<enum IMCObjectName, int> PDOmap::map(int slaveIndex, enum dataDirection direction)
```
Olavhaasie: The `enum` keyword is not necessary here

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -57,24 +59,29 @@ enum class IMCObjectName
...
-  void addObject(IMCObjectName objectname);
+  void addObject(IMCObjectName object_name);
   std::map<IMCObjectName, int> map(int slaveIndex, dataDirection direction);
 
 private:
-  void initAllObjects();
+  /** This function is used to sort the objects in the all_objects according to data length.*/
   void sortPDOObjects();
+
+  /** This function configures the PDO in the IMC using the given base register address and sync manager address.*/
```
Olavhaasie: You should also add what the function returns, i.e. what maps the `IMCObjectName` to?

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -57,24 +59,29 @@ enum class IMCObjectName
...
-  const int nrofRegs = 4;
-  const int objectSizes[3] = { 8, 16, 32 };
+  std::map<IMCObjectName, IMCObject> PDO_objects;
+  std::map<IMCObjectName, IMCObject> all_objects;
+  std::vector<std::pair<IMCObjectName, IMCObject>> sorted_PDO_objects;
+  std::map<IMCObjectName, int> byte_offsets;
+
+  const int bits_per_register = 64;           // Maximum amount of bits that can be constructed in one PDO message.
+  const int nr_of_regs = 4;                   // Amount of registers available.
+  const int object_sizes[3] = { 32, 16, 8 };  // Available sizes.
```
Olavhaasie, line 84: Why did you reverse this?

march_hardware/src/PDOmap.cpp
```diff
@@ -8,125 +8,142 @@ namespace march4cpp
...
+
+  if (totalBits > this->nr_of_regs * this->bits_per_register)
   {
-    ROS_FATAL("Too many objects in PDO Map (total bits %d, only %d allowed)", totalbits,
-              this->nrofRegs * this->bitsPerReg);
+    ROS_FATAL("Too many objects in PDO Map (total bits %d, only %d allowed)", totalBits,
+              this->nr_of_regs * this->bits_per_register);
     throw std::exception();
   }
 }
```
Olavhaasie: I would move this check to the `addObject` so you know which object you added was too much

march_hardware/src/PDOmap.cpp
```diff
@@ -8,125 +8,142 @@ namespace march4cpp
...
-    sdo_bit16(slaveIndex, SMAddress, count, 0x1600);
+    for (int unusedRegister = currentReg; unusedRegister < baseRegister + this->nr_of_regs; unusedRegister++)
+    {
+      sdo_bit8(slaveIndex, unusedRegister, 0, 0);
+    }
   }
-  sdo_bit8(slaveIndex, SMAddress, 0, count);
-  return this->byteOffsets;
+
+  // Active the sync manager again
```
Olavhaasie:
```suggestion
  // Activate the sync manager again
```

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -57,24 +59,29 @@ enum class IMCObjectName
...
-  std::vector<std::pair<IMCObjectName, IMCObject>> sortedPDOObjects;
-  std::map<IMCObjectName, int> byteOffsets;
 
-  const int bitsPerReg = 64;
-  const int nrofRegs = 4;
-  const int objectSizes[3] = { 8, 16, 32 };
+  std::map<IMCObjectName, IMCObject> PDO_objects;
+  std::map<IMCObjectName, IMCObject> all_objects;
+  std::vector<std::pair<IMCObjectName, IMCObject>> sorted_PDO_objects;
+  std::map<IMCObjectName, int> byte_offsets;
```
Olavhaasie: This member variable is also local to `configurePDO` and can be removed

march_hardware/src/PDOmap.cpp
```diff
@@ -8,125 +8,142 @@ namespace march4cpp
...
+  this->all_objects[IMCObjectName::QuickStopOption] = IMCObject(0x605A, 16);
 }
 
-void PDOmap::addObject(IMCObjectName objectname)
+void PDOmap::addObject(IMCObjectName object_name)
 {
-  if (this->allObjects.count(objectname) != 1)
+  if (this->all_objects.count(object_name) != 1)
   {
     ROS_WARN("IMC object does not exist (yet), or multiple exist");
```
Olavhaasie: It is not possible for a map to have multiple keys, so this warning message is misleading

march_hardware/src/PDOmap.cpp
```diff
@@ -8,125 +8,142 @@ namespace march4cpp
...
+  this->all_objects[IMCObjectName::TargetPosition] = IMCObject(0x607A, 32);
+  this->all_objects[IMCObjectName::TargetTorque] = IMCObject(0x6071, 16);
+  this->all_objects[IMCObjectName::QuickStopDeceleration] = IMCObject(0x6085, 32);
+  this->all_objects[IMCObjectName::QuickStopOption] = IMCObject(0x605A, 16);
 }
 
-void PDOmap::addObject(IMCObjectName objectname)
+void PDOmap::addObject(IMCObjectName object_name)
 {
-  if (this->allObjects.count(objectname) != 1)
```
Olavhaasie, line 16: I think this will be easier to read if you replace it with the `find` function from `map`. This will also stop when it has found the element and not continue to count, while there can only exist one in a unique map.
```c++
if (this->all_objects.find(object_name) == this->all_objects.end())
```

march_hardware/src/PDOmap.cpp
```diff
@@ -8,125 +8,142 @@ namespace march4cpp
...
+  if (this->all_objects.count(object_name) != 1)
   {
     ROS_WARN("IMC object does not exist (yet), or multiple exist");
     return;
```
Olavhaasie, line 31: These returns are redundant

march_hardware/src/PDOmap.cpp
```diff
@@ -8,125 +8,142 @@ namespace march4cpp
...
   }
-  else if (this->PDOObjects.count(objectname) != 0)
+  else if (this->PDO_objects.count(object_name) != 0)
   {
     ROS_WARN("IMC object is already added to PDO map");
     return;
```
Olavhaasie, line 37: This one as well

Olavhaasie (changes requested):
I have also made a lot of comments on code you haven't touched, but I think should be changed anyway (because this is the **rewrite**)

march_hardware/include/march_hardware/PDOmap.h
Joris-JW: Because in the previous code the object size were sorted from small to large and when using the sortPDOObjects function they would use a while loop to iterate over all the PDO's and removing the last one (which is the largest of the list) until no object was left.

I removed this while loop and inserted a foreach loop which is way nicer, but it is vital to map from largest to smallest. So the bitsizes should be sorted from 32 to 16 to 8.

march_hardware/include/march_hardware/PDOmap.h
Joris-JW: I think I added what u are referring to. However could not find it in the style guide, let me know if u approve.

march_hardware/include/march_hardware/PDOmap.h
Joris-JW: I think I added what u are referring to. However could not find it in the style guide, let me know if u approve.

march_hardware/include/march_hardware/PDOmap.h
Joris-JW: nice, didn't know about this.

march_hardware/src/PDOmap.cpp
Joris-JW: not really sure why this is misleading, if the given item is already in the pdo_objects map it should not be added anymore

march_hardware/src/PDOmap.cpp
Joris-JW: fixed

march_hardware/src/PDOmap.cpp
Joris-JW: I have added the position integer because otherwise it will be a lot more code ):

march_hardware/src/PDOmap.cpp
Joris-JW: fixed

march_hardware/src/PDOmap.cpp
Joris-JW: I agree that it is not really common to not define such vital variables as constants of the PDOmap namespace. However in this case the classes are well commented and defining these variables as constants would result in a lot more code and not much more readability, maybe even less.

BUT!! I am a fan of following style guided rules so if u (or other reviewers) think we should name these numbers as constants, will do.

march_hardware/src/PDOmap.cpp
```diff
@@ -6,129 +6,152 @@
 
 namespace march4cpp
 {
-PDOmap::PDOmap()
+std::unordered_map<IMCObjectName, IMCObject> PDOmap::all_objects =
 {
-  this->initAllObjects();
-}
+  { IMCObjectName::StatusWord, IMCObject(0x6041, 16) },
```
RPKrijnen, line 10: If you're rewriting the mapping anyway, please add 0x6007 the object in which we can set the way we react to EtherCAT communication loss

march_hardware/src/PDOmap.cpp
```diff
@@ -6,129 +6,151 @@
...
+};
 
-void PDOmap::addObject(IMCObjectName objectname)
+void PDOmap::add_object(IMCObjectName object_name)
 {
-  if (this->allObjects.count(objectname) != 1)
+  if (PDOmap::all_objects.find(object_name) == PDOmap::all_objects.end())
   {
-    ROS_WARN("IMC object does not exist (yet), or multiple exist");
+    ROS_WARN("IMC does not exists in in initialised objects so can not be added to PDO");
```
martijnvandermarel, line 30: Strange sentence.

march_hardware/src/PDOmap.cpp
```diff
@@ -6,129 +6,151 @@
...
     return;
   }
-  else if (this->PDOObjects.count(objectname) != 0)
+
+  if (this->PDO_objects.count(object_name) != 0)
   {
-    ROS_WARN("IMC object is already added to PDO map");
+    ROS_WARN("IMC object %i is already added to PDO map", object_name);
```
martijnvandermarel, line 36: You will probably want to print the hex value or the name of the object, right? This would print a (decimal) integer?

march_hardware/src/PDOmap.cpp
```diff
@@ -6,129 +6,151 @@
...
+  std::map<IMCObjectName, int> byte_offsets;
+
+  sdo_bit8(slaveIndex, baseRegister, 0, 0);
+
+  for (const auto& nextObject : sorted_PDO_objects)
   {
-    // Check if register is still empty
-    if (sizeleft == this->bitsPerReg)
+    sizeLeft -= nextObject.second.length;
+    if (sizeLeft < 0)
```
martijnvandermarel: What if sizeLeft equals 0?

march_hardware/src/PDOmap.cpp
```diff
@@ -6,129 +6,151 @@
...
+  sdo_bit8(slaveIndex, baseSyncManager, 0, 0);
+  int currentPDONr = (currentReg - baseRegister) + 1;
+  sdo_bit16(slaveIndex, baseSyncManager, currentPDONr, currentReg);
+
+  // Explicitly disable PDO registers which are not used
+  currentReg++;
+  if (currentReg <= (baseRegister + nr_of_regs))
   {
-    count++;
-    sdo_bit16(slaveIndex, SMAddress, count, 0x1600);
```
martijnvandermarel, line 101: The 0x1600 here was probably originally a bug? I can't imagine that should have been there. Should probably have been startReg or something.

march_hardware/src/PDOmap.cpp
```diff
@@ -6,129 +6,151 @@
...
-  // For the unused registers, set count to zero
-  for (int i = reg; i < startReg + this->nrofRegs; i++)
-  {
-    sdo_bit32(slaveIndex, i, 0, 0);
-  }
-  // For all filled registers, set data to Sync Manager object
-  int count = 0;
-  for (int i = startReg; i <= lastFilledReg; i++)
+
+  // Make sure the last PDO and sync manager are activated
```
martijnvandermarel: This comment is not really accurate for the 4 lines of code below. The PDO is enabled, but the sync manager is disabled, then updated (but not enabled again).

march_hardware/src/PDOmap.cpp
```diff
@@ -6,129 +6,151 @@
...
+  sdo_bit8(slaveIndex, currentReg, 0, counter - 1);
+  sdo_bit8(slaveIndex, baseSyncManager, 0, 0);
+  int currentPDONr = (currentReg - baseRegister) + 1;
+  sdo_bit16(slaveIndex, baseSyncManager, currentPDONr, currentReg);
+
+  // Explicitly disable PDO registers which are not used
+  currentReg++;
+  if (currentReg <= (baseRegister + nr_of_regs))
```
martijnvandermarel: I am breaking my head a bit thinking of off-by-one errors here. I think here it will work out, but it could be better I think if the if condition were changed to a less than. Say you use all four registers (say 0x1600-0x1603). Then the currentReg will be incremented to 0x1604. The if condition is true (equality). However the for loop will quit immediately since unusedRegister (=0x1604) is not smaller than baseRegister+nr_of_regs (= also 0x1604). So that could probably be improved.

martijnvandermarel (commented):
I added a few comments. I think it looks good in general. The order of sdo messages is sometimes different than the original, but I looked at it with the CoE programming manual and I think this should work fine.

march_hardware/src/PDOmap.cpp
Joris-JW: indeed the object name would be best. But currently these names are structured in an enum and to print these values a switch case statement is required for name. I thought it would not improve the message critically since we can cross-check the integer with the enum list in the code. If u (or other reviews) do not agree with this structure I will create a more suitable solution.

march_hardware/src/PDOmap.cpp
Joris-JW: Changed to a more suitable message.

march_hardware/src/PDOmap.cpp
Joris-JW: If sizeLeft is 0 then the object still fits in the currently selected PDO.

march_hardware/src/PDOmap.cpp
Joris-JW: split into multiple comments

march_hardware/src/PDOmap.cpp
Joris-JW: Indeed a bug, maybe  when this code was used only one PDO was mapped therefore only 0x1600 was written to the sync manager. Anyway, it is removed.

march_hardware/src/PDOmap.cpp
Joris-JW: hmm I see what u are implying. I think it would be indeed more logical to have the same condition in the if statement as in the for loop such as:
```
for (int unusedRegister = currentReg; unusedRegister <= baseRegister + this->nr_of_regs; unusedRegister++)
```
In this case when the currentReg is increment to an new unused register like 1604, the for loop will loop one time to indeed disable this PDO.

march_hardware/src/PDOmap.cpp
```diff
@@ -124,7 +126,7 @@ std::map<IMCObjectName, int> PDOmap::configurePDO(int slaveIndex, int baseRegist
   currentReg++;
   if (currentReg <= (baseRegister + nr_of_regs))
   {
-    for (int unusedRegister = currentReg; unusedRegister < baseRegister + this->nr_of_regs; unusedRegister++)
+    for (int unusedRegister = currentReg; unusedRegister <= (baseRegister + this->nr_of_regs); unusedRegister++)
```
martijnvandermarel: I think the if and the for loop conditions should both be less than (not equal). Now you will write sdo messages for unused registers up to 0x1604, while the PDO registers only go up till 0x1603.

martijnvandermarel (dismissed):

march_hardware/src/PDOmap.cpp
Olavhaasie: Where is this documented?

march_hardware/src/PDOmap.cpp
Olavhaasie: What I meant was that the final addition of `, or multiple exist` should be removed, since that is not possible in a unique map.
```suggestion
     ROS_WARN("IMC object does not exist (yet)");
```

march_hardware/src/PDOmap.cpp
Joris-JW: https://google.github.io/styleguide/cppguide.html#Constant_Names

march_hardware/src/PDOmap.cpp
martijnvandermarel: Ah yes I see. It looks a bit counterintuitive, but if you think it's ok that's fine.

march_hardware/src/PDOmap.cpp
martijnvandermarel: Ah I see. It's for you to decide as a team since you'll be using it, not me ;)

march_hardware/src/PDOmap.cpp
martijnvandermarel: See my comment in my later review.

march_hardware/src/IMotionCube.cpp
```diff
@@ -76,41 +76,38 @@ void IMotionCube::validateMosiPDOs()
...
+  int abort_con = sdo_bit16(slaveIndex, 0x6007, 0, 1);
 
   // set the ethercat rate of encoder in form x*10^y
-  success &= sdo_bit8(slaveIndex, 0x60C2, 1, ecatCycleTime);
-  success &= sdo_bit8(slaveIndex, 0x60C2, 2, -3);
+  int rate_ec_x = sdo_bit8(slaveIndex, 0x60C2, 1, ecatCycleTime);
+  int rate_ec_y = sdo_bit8(slaveIndex, 0x60C2, 2, -3);
 
-  ROS_ASSERT_MSG(success, "Writing initial settings to IMC %d failed", this->slaveIndex);
+  if ((mode_of_op && max_pos_lim && min_pos_lim && stop_opt && stop_decl && abort_con && rate_ec_x && rate_ec_y) == 0)
```
Olavhaasie: An `&&` operand will return a `bool`, however you check the result using `== 0`, which does not seem valid (but may work).
```suggestion
  if (mode_of_op && max_pos_lim && min_pos_lim && stop_opt && stop_decl && abort_con && rate_ec_x && rate_ec_y)
```

march_hardware/src/PDOmap.cpp
```diff
@@ -6,129 +6,156 @@
...
+  return byte_offsets;
 }
 
-void PDOmap::sortPDOObjects()
+std::vector<std::pair<IMCObjectName, IMCObject>> PDOmap::sortPDOObjects()
 {
-  // Sort from small to large
-  int totalbits = 0;
-  for (int i = 0; i < (sizeof(this->objectSizes) / sizeof(this->objectSizes[0])); i++)
+  int totalBits = 0;
```
Olavhaasie: the `totalBits` variable is no longer used in this function

Olavhaasie (changes requested):
Final comments I promise :innocent:

march_hardware/src/PDOmap.cpp
Olavhaasie: What I meant was where are these numbers documented? Why are they `0x1600` and `0x1C12`?

Joris-JW: IMC manual paragraph 2.3; **2.3 Objects that define SDOs and PDOs**

Olavhaasie: Ah okay I see

Olavhaasie (dismissed):
Nice work, let's see if it works :+1: :100:

Olavhaasie (dismissed):
Apparently it did not work, let's see if it works now :slightly_smiling_face:

RPKrijnen (dismissed):
Very nice!

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -76,7 +76,7 @@ class PDOmap
    * @return combination of both address, sub-index and length (Example control word: 60400010h.) */
   static uint32_t combineAddressLength(uint16_t address, uint16_t length);
 
-  std::unordered_map<IMCObjectName, IMCObject> PDO_objects;
+  std::map<IMCObjectName, IMCObject> PDO_objects;
```
Olavhaasie, line 80: Why did you revert it to a `map`?

Olavhaasie (changes requested):
The naming is a lot better now, thanks

march_hardware/include/march_hardware/PDOmap.h
Joris-JW: this is necessary to make it work (see commit message)

march_hardware/src/PDOmap.cpp
```diff
@@ -6,129 +6,153 @@
...
-    // Get next object (from end, because sorted from small to large)
-    std::pair<IMCObjectName, IMCObject> nextObject = this->sortedPDOObjects.back();
-    this->sortedPDOObjects.pop_back();
-    // Add next object to map
+
+    int byteOffset = (bits_per_register - (size_left + nextObject.second.length)) / 8;
+    byte_offsets[nextObject.first] = byteOffset;
+
+    sdo_bit32(slave_index, current_register, counter,
+              PDOmap::combineAddressLength(nextObject.second.address, nextObject.second.length));
```
Olavhaasie: This method only applies to the `IMCObject` class, so I think it is better to make this a member function of the class. Then you call it with `nextObject.second.combineAddressLength()`. Furthermore, the method signature uses `uint16_t` while the members are of type `int`. So change the members' type too (see other comment).

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -6,36 +6,35 @@
...
 
 #include <march_hardware/EtherCAT/EthercatSDO.h>
 
 namespace march4cpp
 {
+/** Store IMC data as a struct to prevent data overlap.*/
 struct IMCObject
 {
   int address;  // in IMC memory (see IMC manual)
   int length;   // bits (see IMC manual)
```
Olavhaasie: These should be `uint16_t` and `const` to prevent any bugs by editing these values in the algorithm

Olavhaasie (changes requested):

march_hardware/include/march_hardware/PDOmap.h
Joris-JW: It is possible naming the variable as **uint16_t** (although clion gives a warning because these are not initialized). However, declaring the variables **const** would mean they would directly need to be initialized which is not possible because the variables vary per object.

Olavhaasie: Ah yes `const` is not possible, since the other constructor still exists. But if you initialize the variables in the initializer list then can be const since you never want to edit it.

When I tried removing the constructor it errors on https://github.com/project-march/hardware-interface/blob/b0373bd305d7371fdb2d4e53e9291d47ca8e73a0/march_hardware/src/PDOmap.cpp#L40

For some reason this line uses the default constructor, so let's leave the `const` for now, but still change the type

march_hardware/src/PDOmap.cpp
```diff
@@ -6,129 +6,153 @@
...
-    ROS_WARN("IMC object is already added to PDO map");
+    ROS_WARN("IMC object %i is already added to PDO map", object_name);
     return;
   }
-  else
+
+  this->PDO_objects.insert({ object_name, PDOmap::all_objects[object_name] });  // NOLINT(whitespace/braces)
+
+  int total_used_bits = 0;
+  for (std::pair<IMCObjectName, IMCObject> PDO_object : this->PDO_objects)
```
Olavhaasie: You should use a constant reference, to avoid copying the pair in each loop.
```suggestion
  for (const std::pair<IMCObjectName, IMCObject>& PDO_object : this->PDO_objects)
```

Olavhaasie (changes requested):

march_hardware/include/march_hardware/PDOmap.h
Joris-JW: I've encountered the same problem. weird though. I'll leave it like it is.

Ciliaclaij (approved):
Yasssss

Olavhaasie (approved):

PR #193, opened 2020-01-08
Fix/pm 196 fix return value initation imc check

Joris-JW:
Closes PM-196

## Description
Whenever a SDO message is send using the etherCAT protocol there is a possibility that the message is not fully received or processed. In this case the working counter, which is the return value of an SDO message, remains zero. So if this happens within the code the conclusion is that the message has not been processed by the IMC (slave).  This PR migrates the working counter check to the general SDO send methods. This should improve liability

## Changes
* Updated the SDO send functions to check if the working counter is 0
* Removed the check within the IMC initialization functions
* Removed the (unused) factor group settings 

## Tested
* Tested on rotational test joint
* Tested on exo


Olavhaasie:
This change has been merged to #192 since that fix was necessary for this to work

PR #197, opened 2020-01-22
Fix: pdo map register

Olavhaasie:
## Description
Fixes a bug introduced in #192 where the register offset was not added to the byte offset for registers higher than 0.

## Changes
* Added register offset to byte offset
* Changed PDO `map` to `unordered_map`

### DIFF
--- a/march_hardware/include/march_hardware/IMotionCube.h
+++ b/march_hardware/include/march_hardware/IMotionCube.h
@@ -3,7 +3,7 @@
 #ifndef MARCH_HARDWARE_IMOTIONCUBE_H
 #define MARCH_HARDWARE_IMOTIONCUBE_H
 
-#include <map>
+#include <unordered_map>
 #include <string>
 
 #include <march_hardware/ActuationMode.h>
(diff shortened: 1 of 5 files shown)


RutgerVanBeek (approved):

march_hardware/include/march_hardware/IMotionCube.h
```diff
@@ -3,7 +3,7 @@
 #ifndef MARCH_HARDWARE_IMOTIONCUBE_H
 #define MARCH_HARDWARE_IMOTIONCUBE_H
 
-#include <map>
+#include <unordered_map>
```
RPKrijnen, line 6: Did @JorisWeeda not notice that it mattered in what order the map was made? Or do your changes circumvent this issue?

RutgerVanBeek: This bug was not noticed because when an ordered map is used all the vital components are in the first register. When an unordered map is used the bug would actually show up, because essential components would end up in the second register and would be read wrongly.

Joris-JW (approved):

PR #201, opened 2020-01-28
Fix: incorrect use of encoders during startup

Olavhaasie:
## Description
In the CoE manual @Deakensbubbles found that you should not read out encoders before the IMC is in the `SWITCHED_ON` state. So, we (re)moved any lines that did this before the IMC was in that state.

## Changes
* Remove printing encoder values before and after resetting IMC
* Move setting controller commands after the IMC is operational

### DIFF
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -514,7 +514,6 @@ bool IMotionCube::goToOperationEnabled()
 
   if (this->actuationMode == ActuationMode::position)
   {
-    angleRead = this->getAngleIU();
     this->actuateIU(angleRead);
   }
   if (this->actuationMode == ActuationMode::torque)
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -97,18 +97,6 @@ void IMotionCube::writeInitialSettings(uint8 ecatCycleTime)
   // Quick stop deceleration
   int stop_decl = sdo_bit32(slaveIndex, 0x6085, 0, 0x7FFFFFFF);
 
-  // No FUCKING HOMING
-  int no_hom_pos = sdo_bit8(slaveIndex, 0x6098, 0, 0);
-
-  // No FUCKING HOMING
-  int no_hom_vel1 = sdo_bit32(slaveIndex, 0x6099, 1, 0);
-
-  // No FUCKING HOMING
-  int no_hom_vel2 = sdo_bit32(slaveIndex, 0x6099, 2, 0);
-
-  // No FUCKING HOMING
-  int no_hom_acc = sdo_bit32(slaveIndex, 0x609A, 0, 0);
-
   // Abort connection option code
   int abort_con = sdo_bit16(slaveIndex, 0x6007, 0, 1);
 
@@ -116,8 +104,7 @@ void IMotionCube::writeInitialSettings(uint8 ecatCycleTime)
   int rate_ec_x = sdo_bit8(slaveIndex, 0x60C2, 1, ecatCycleTime);
   int rate_ec_y = sdo_bit8(slaveIndex, 0x60C2, 2, -3);
 
-  if (!(mode_of_op && max_pos_lim && min_pos_lim && stop_opt && stop_decl && abort_con && rate_ec_x && rate_ec_y &&
-        no_hom_pos && no_hom_vel1 && no_hom_vel2 && no_hom_acc))
+  if (!(mode_of_op && max_pos_lim && min_pos_lim && stop_opt && stop_decl && abort_con && rate_ec_x && rate_ec_y))
   {
     ROS_ERROR("Failed writing initial settings to IMC of slave %i", slaveIndex);
   }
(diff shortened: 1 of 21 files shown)


march_hardware/src/IMotionCube.cpp
```diff
@@ -506,18 +511,16 @@ bool IMotionCube::goToOperationEnabled()
...
   {
-    if (this->actuationMode == ActuationMode::position)
-    {
-      this->actuateIU(angleRead);
-    }
-    if (this->actuationMode == ActuationMode::torque)
-    {
-      this->actuateTorque(0);
-    }
+    angleRead = this->getAngleIU();
```
Roelemans: Why do this again?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -86,13 +86,6 @@ void MarchHardwareInterface::init()
 
   initiateIMC();
 
-  // Print all joint positions on startup in case initialization fails.
-  this->read();
-  for (int i = 0; i < num_joints_; ++i)
-  {
-    ROS_DEBUG("[%s] First read position: %f", joint_names_[i].c_str(), joint_position_[i]);
-  }
-
```
Roelemans, line 95: Kan dit niet best handig zijn? Of komt dit sowieso al in de rosbag

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -205,6 +184,20 @@ void MarchHardwareInterface::init()
         }
       }
       joint.prepareActuation();
```
Roelemans, line 186: Dit zet de IMC in de SWITCH_ON state?

Roelemans (commented):
Has this version been tested on exo? Because we tried a lot of versions that did not work.

march_hardware_interface/src/march_hardware_interface.cpp
RPKrijnen: This was the whole point of the PR. When these lines are executed, the iMC is not in SWITCHED_ON state and it is therefore not correct to use the read function.

march_hardware_interface/src/march_hardware_interface.cpp
RPKrijnen: Yes, this function in turn calls the iMotionCube.goToOperationEnabled() function wherein we set the iMC to SWICTHED_ON and then to OPERATION_ENABLE.

march_hardware/src/IMotionCube.cpp
RPKrijnen: I guess you refer to the use in line 500 as well? Because the state of the encoder can change at any instant between the lines 500 and 517 it is best practice to read out the state again. This prevents unnecessary errors.

RPKrijnen (dismissed):
I like these changes :)

march_hardware/src/IMotionCube.cpp
Roelemans: Seems to me like overkill considering the program only has to calculate three if-statements in between 500 and 517 (because the code will error if any of the two above if statements is true). @Olavhaasie Opinion?

RPKrijnen: I don't see how this would hurt our performance. And if we remove the extra check we run the slight risk of not sending the actual current angle to the joint. We'd be taking unnecessary risks, however slight, with no consequence to the functioning of our system.

Olavhaasie: I think this is indeed uneccesary

Roelemans: It is better not to put in a large number of unnecessary checks, it makes code longer than necessary and our descendants may think "no idea why this is here but it might be important so I don't dare remove it". 

Besides, I think that if you are talking about the risk of the encoder value changing, I think the bigger problem will appear if the encoder resets within that 0.1 nanoseconds, because then the motor will try to move to position 0.

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -323,30 +316,12 @@ void MarchHardwareInterface::initiateIMC()
...
-  for (const std::string& joint_name : joint_names_)
-  {
-    Joint joint = marchRobot.getJoint(joint_name);
-
-    if (LOWER_BOUNDARY_ANGLE_IU <= joint.getAngleIU() && joint.getAngleIU() <= UPPER_BOUNDARY_ANGLE_IU)
-    {
-      ROS_WARN("After reset joint: [%s] has angle-value of: %i. Which is within boundary of lower: %i and upper: %i",
-               joint_name.c_str(), joint.getAngleIU(), LOWER_BOUNDARY_ANGLE_IU, UPPER_BOUNDARY_ANGLE_IU);
-    }
-  }
```
Joris-JW, line 349: If after the reset the angles are still a non valid value shouldn't there be an error? I know this was not implemented before but maybe we should?

Joris-JW (commented):
Has this been tested on exo...?

march_hardware_interface/src/march_hardware_interface.cpp
RPKrijnen: We have removed these lines, because you'd be reading out encoder values before the iMC is in SWITCHED_ON state.

Roelemans (approved):

RPKrijnen (approved):
Good job!

PR #202, opened 2020-01-28
Feature: custom imc exceptions

Olavhaasie:
Closes PM-242

## Description
This PR adds new custom `HardwareException` to the `IMotionCube` class. And I have included an `ErrorType` class that specifies the different errors possible. These types are given to the exception when it is thrown. Furthermore, the motion error parsing has been refactored out of the `IMotionCube` class and moved to the new `error` namespace. The `HardwareException` and `ErrorType` classes can be reused for every class in `march_hardware` that throws exceptions.

All the `ROS_ASSERT`s have been replaced by throwing this new exception. The assert statements that check if the PDO mapping have been added have been completely removed. This has been replaced by accessing the elements in the map with `.at()`. This method throws an exception when the key does not exist and should only be exceptional, since it indicates a bug.

## Changes
* Create `HardwareException` class that inherits from `std::exception`
* Create `ErrorType` enum class that lists different types of errors
* Split motion error parsing from `IMotionCube`
* Replace `ROS_ASSERT`s with throwing `HardwareException`
* Add tests for newly added classes
* Fix `march_hardware` tests by using `rosunit` instead of `rostest`


march_hardware/src/IMotionCube.cpp
```diff
@@ -1,95 +1,76 @@
...
-
-// Checks if the compulsory MOSI PDO objects are mapped
-void IMotionCube::validateMosiPDOs()
-{
-  ROS_ASSERT_MSG(this->mosiByteOffsets.count(IMCObjectName::ControlWord) == 1, "ControlWord not mapped");
+  PDOmap map_mosi = PDOmap();
+  map_mosi.addObject(IMCObjectName::ControlWord);  // Compulsory!
+  map_mosi.addObject(IMCObjectName::TargetPosition);
+  map_mosi.addObject(IMCObjectName::TargetTorque);
+  this->mosi_byte_offsets_ = map_mosi.map(this->slaveIndex, DataDirection::MOSI);
```
RutgerVanBeek, line 58: Why is the validation removed?

march_hardware/src/IMotionCube.cpp
```diff
@@ -110,409 +91,182 @@ void IMotionCube::writeInitialSettings(uint8 ecatCycleTime)
...
-  return this->encoder.getAngleIU(this->misoByteOffsets[IMCObjectName::ActualPosition]);
+  return this->encoder_.getAngleIU(this->miso_byte_offsets_.at(IMCObjectName::ActualPosition));
 }
 
-uint16 IMotionCube::getStatusWord()
+uint16_t IMotionCube::getStatusWord()
 {
-  ROS_ASSERT_MSG(this->misoByteOffsets.count(IMCObjectName::StatusWord) == 1, "StatusWord not defined in PDO "
-                                                                              "mapping, so can't get status word");
-  return get_input_bit16(this->slaveIndex, this->misoByteOffsets[IMCObjectName::StatusWord]).ui;
```
RutgerVanBeek, line 207: Why remove all the warnings? This can help with debugging. Same for the others ofcourse.

march_hardware/src/error/motion_error.cpp
```diff
@@ -0,0 +1,67 @@
...
+const char* DETAILED_MOTION_ERRORS[DETAILED_MOTION_ERRORS_SIZE] = {
+  "TML stack overflow. ",
+  "TML stack underflow. ",
+  "Homing not available. ",
+  "Function not available. ",
+  "UPD ignored. ",
+  "Cancelable call ignored. ",
+  "Positive software limit switch is active. ",
+  "Negative software limit switch is active. ",
+  "Invalid S-curve profile. ",
```
RutgerVanBeek, line 36: There are 15 or 16 I believe. I think we should add them all.

march_hardware_builder/test/yaml/joint/joint_correct_position_mode.yaml
```diff
@@ -1,2 +1,10 @@
 actuationMode: position
 allowActuation: false
+imotioncube:
+  slaveIndex: 1
+  encoder:
+    resolution: 16
+    minPositionIU: 22134
+    maxPositionIU: 43436
+    zeroPositionIU: 24515
+    safetyMarginRad: 0.05
```
RutgerVanBeek, line 10: Why is this recalibration part of this PR?

RutgerVanBeek (changes requested):

Joris-JW (commented):
Nice! very fond of custom errors. Since this is difficult to verify  first do some basic testing on the exoskeleton before merging maybe?

march_hardware_builder/test/yaml/joint/joint_correct_position_mode.yaml
Olavhaasie: This is a test file that I had to change. Has no effect on the functioning of the real stuff

march_hardware/src/error/motion_error.cpp
Olavhaasie: I also noticed this but I was only able to find these 9 in the CoE manual. I believe it says the higher bits are reserved

march_hardware/src/IMotionCube.cpp
Olavhaasie: See commit https://github.com/project-march/hardware-interface/pull/202/commits/26d64d63dfaee741bdc53ba444e3e5780d453236

> This is replaced by using the `.at()` method on the maps. This method throws an exception when the key does not exist.

The asserts were not warnings and would cause the program to abort abruptly, which is not nice to do and cannot be caught. On the other hand, this is something you can not recover from, since this should never happen.

march_hardware/src/IMotionCube.cpp
Olavhaasie: Since they did the same as the other `ASSERT`s, which were also removed (see other comment)

RutgerVanBeek (dismissed):

RutgerVanBeek (dismissed):
First test on exo of course

Joris-JW (dismissed):

Joris-JW (dismissed):

Roelemans (approved):
Sure

Joris-JW (approved):
sick

PR #206, opened 2020-01-31
Feature: add coverage report

Olavhaasie:
Closes PM-79

## Description
This PR adds coverage reports uploading to [codecov.io](https://codecov.io/gh/project-march/hardware-interface/branch/feature%2FPM-79-add-coverage-report). The test coverage is also uploaded but under a different flag and is not reported in normal project statistics.

You can also generate these reports locally by running:

    $ colcon build --cmake-args -DENABLE_COVERAGE_TESTING=ON -DCMAKE_BUILD_TYPE=Debug --cmake-target coverage_report --cmake-target-skip-unavailable 

This will build all packages with coverage flags enabled, run tests and create coverage reports. If you want to find the generated report of a package you can go to `build/<package>/coverage_report` and open `index.html` with your browser.

image

## Changes
* Enable building with coverage enabled on Travis
* Add codecov uploading scripts to Travis
* Fix usage of `code_coverage` package, so that you get no more errors saying that the build is optimized, which is not good for coverage.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,17 @@
+coverage:
+  status:
+    project:
+      default: off
+      production:
+        target: auto
+        flags:
+          - production
 fixes:
   - "src/hardware-interface/::"
+flags:
+  production:
+    paths:
+      - "*/src/**/*"
+  test:
+    paths:
+      - "*/test/**/*"
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+fixes:
+  - "src/hardware-interface/::"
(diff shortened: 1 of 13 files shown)


RutgerVanBeek (approved):

codecov.yml
```diff
@@ -0,0 +1,21 @@
+coverage:
+  status:
+    project:
+      default: off
+      production:
+        target: auto
```
Joris-JW, line 6: (thought I commented this) why use **target: auto** and not a new, call it MARCH, coverage standard for projects with code coverage?

Joris-JW (commented):
lost comment ):

codecov.yml
Joris-JW: Same question in other PR so this one is redundant.

Joris-JW (approved):

PR #209, opened 2020-02-05
Fix random high effort peaks bug

Roelemans:
## Description

Fix for the bug that the effort sometimes gets set to a very high value. 

The problem was the following:
There was a loop over all joints that first multiplies the effort of joint i by 1000, then enforces limits **on all joints** and then sends the effort joint i to the IMC. Hence, the effort limit got enforced on joints where the effort was not yet multiplied by 1000. The problem is that sometimes the minimum effort is nonzero, when the joint gets close to its limits. Then the effort limit thinks "hey the minimum effort is 1500 and it is only 4 so imma set it to 1500.", then when the loop get to this joint, the effort is multiplied by 1000 for a total of 1500000, which then gets enforced back to its max limit again by the effort limits enfocer.

Fixed by first multiplying all joint efforts by 1000 then enforcing limits on all joints only once.

### DIFF
--- a/march_hardware_interface/src/march_hardware_interface.cpp
+++ b/march_hardware_interface/src/march_hardware_interface.cpp
@@ -263,9 +263,23 @@ void MarchHardwareInterface::read(const ros::Duration& elapsed_time)
 
 void MarchHardwareInterface::write(const ros::Duration& elapsed_time)
 {
-  joint_effort_command_copy.clear();
-  joint_effort_command_copy.resize(joint_effort_command_.size());
-  joint_effort_command_copy = joint_effort_command_;
+  if (joint.getActuationMode() == march::ActuationMode::position)
+  {
+    positionJointSoftLimitsInterface.enforceLimits(elapsed_time);
+  }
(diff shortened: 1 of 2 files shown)


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -263,10 +263,21 @@ void MarchHardwareInterface::read(const ros::Duration& elapsed_time)
 
 void MarchHardwareInterface::write(const ros::Duration& elapsed_time)
 {
+  for (int i = 0; i < num_joints_; i++)
+  {
+    // Enlarge joint_effort_command so dynamic reconfigure can be used inside it's bounds
```
Ishadijcks:
```suggestion
    // Enlarge joint_effort_command so dynamic reconfigure can be used inside its bounds
```

Bart-van-Ingen (commented):
Nice! The explanation makes sense, I am excited to see if this will solve the problem.

Olavhaasie (approved):

Joris-JW (approved):

PR #212, opened 2020-02-10
Fix: incorrect types

Olavhaasie:
Closes PM-261

## Description
The hardware (interface) was not using correct types in some places for the control word, status word, torque and position radians. I changed all types that were `float` to `double` since that is the type used by ros_control.

## Changes
* Make types consistent for:
  * control word (`uint16_t`)
  * status word (`uint16_t`)
  * torque (`int16_t`)
  * IU position (`int32_t`)
  * radians position (`double`)
  * byte offsets (`uint8_t`)
* Change integer types from soem types to c++ built-in types, i.e. `int16` -> `int16_t`


RutgerVanBeek (commented):
Is the change from float to int16_t for torque correct? It seems weird, but you did this consistently.

Olavhaasie:
> Is the change from float to int16_t for torque correct? It seems weird, but you did this consistently.

I also thought this was weird, however the CoE manual suggests it should be a signed 16 bit integer. This has a range of −32,768 through 32,767, which contains our values. For doing calculations with the controller a `double` type is used.

image

image

RutgerVanBeek (approved):

Roelemans (approved):

PR #214, opened 2020-02-11
Fix segmentation fault when launching

Olavhaasie:
## Description
Fixed the issue where the hardware interface would have a segmentation fault on launch. 
In a last PR I changed the `MarchHardwareInterface` to accept a `MarchRobot` as parameter for its constructor. However, this meant the `MarchRobot` had to be moved. It cannot be copied because `EthercatMaster` has a `std::thread` which can not be copied.
However, the `MarchRobot` class used `unique_ptr`s for `EthercatMaster` and `PowerDistributionBoard`. A move would result in the `unique_ptr` being moved which would result in the original `unique_ptr` pointing to nothing. Then when the old `MarchRobot` was destructed it would try to access the moved ptr, which caused the segmentation fault.

I changed the `MarchRobot` to no longer use `unique_ptr` since there was no reason to.

## Changes
* Removed `unique_ptr`s in `MarchRobot`
* Explicitly removed copy constructor and assignments on `MarchRobot` and `EthercatMaster` to make it clear that these can not be used

### DIFF
--- a/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
+++ b/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
@@ -24,7 +24,6 @@ class EthercatMaster
   int expectedWKC;
 
   std::thread EcatThread;
-  std::vector<Joint>* jointListPtr;
 
   int maxSlaveIndex;
   int ecatCycleTimems;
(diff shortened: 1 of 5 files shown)


Olavhaasie:
I recommend reading into [move semantics](https://www.learncpp.com/cpp-tutorial/15-1-intro-to-smart-pointers-move-semantics/) if you like to know more

Roelemans (dismissed):

Roelemans (approved):
Test by Joris, it works!

Joris-JW (approved):

Joris-JW (approved):
Please check if major changes in the hardware interface do not influence the general behavior or ask someone with an ethernet connection to check the changes. Except for that, nice fix.

PR #215, opened 2020-02-12
Feature: move urdf to builder

Olavhaasie:
First part of PM-216

## Description
In order to resolve issue PM-216, the hardware builder needs access to the URDF in order to calculate the zero position from the limits defined in the URDF. Currently the URDF is read by the `MarchHardwareInterface`, which is after the `MarchRobot` has been built. Therefore, I moved the URDF loading to the `HardwareBuilder`. This also opens up other possibilities, for example only allowing actuation when a joint has been defined non fixed, and using the soft limits from the URDF instead of the `safetyMarginRad` field.

## Changes
* Moved URDF loading from `MarchHardwareInterface` to `HardwareBuilder`
* Added URDF member to `MarchRobot`
* Split initializing `joint_names_` and `num_joints_` to constructor, since the URDF is now available there
* Create `reserveMemory` method that reserves memory in the `MarchHardwareInterface`


Ishadijcks:
The reason we never did this is because the `march_hardware` is now dependent on a ROS package. You should consider the consequences of this dependency carefully for future March teams who might not want to use ROS.

RutgerVanBeek (approved):

Roelemans (approved):

Ishadijcks:
Or not...

Roelemans:
We discussed your comment and came up with a few alternatives, but concluded that they are more effort than they are worth. We appreciate your useful input though :)

PR #217, opened 2020-02-18
Feature: add zero position calculation

Olavhaasie:
Closes PM-216

## Description
This PR introduces a refactor on `Encoder` to support calculating the zero position from the limits defined in the URDF. 

The way to calculate the zero position is

    zero_position = lower_limit_iu + lower_limit_rad * iu_per_rad

However, the `iu_per_rad` can be calculated in two different ways. The first is by dividing the total amount of possible positions of the encoder by 2π

    iu_per_rad = (2^(total_positions) - 1) / 2π

the second by dividing the difference of the iu limits by the difference in radian limits.

    iu_per_rad = (upper_limit_iu - lower_limit_iu) / (upper_limit_rad - lower_limit_rad)

Unfortunately, the limits we have defined in the URDF are not exactly equal to what we measure in the encoder units. Therefore, the first could differ from the first. This is especially the case in the linear joints where it is almost 10%. This will be resolved in PM-278. Currently, I have decided to use the first calculation for converting IU to radians and the other way around, since I thought that that calculation would be more exact.

I also had to change a lot of tests and writing new tests. I have removed some tests for the hardware builder that were indirectly testing the encoder functionality (by checking for incorrect resolutions). Furthermore, the march3 and test_joint_rotational tests are failing, since the march3 URDF does not have any defined soft limits and the `test_joint_rotational`'s calibration is way off. So, that needs to be fixed.

## Changes
* Refactor `Encoder` class to accept limits and calculate zero position
* Added warning when IU per radians calculated from encoder limits differs by 10% from IU per radians calculated from encoder positions.
* Added `ErrorType` when hard and soft limits are conflicting
* Added `ErrorType` when the given encoder resolution is out of range
* Wrote tests for `Encoder`
* Removed some tests in `march_hardware_builder` package that were redundant and/or testing functionality of the `Encoder`, which is now tested
* Removed `zerPositionIU` and `safetyMarginRad` from robot yamls, since these are now calculated from the URDF


march_hardware/include/march_hardware/Encoder.h
```diff
@@ -9,47 +9,40 @@ namespace march
...
-  int32_t getAngleIU(uint8_t ActualPositionByteOffset);
+  /*
+   * Converts encoder Incremental Units to radians.
+   */
+  double toRad(int32_t iu) const;
 
-  double IUtoRad(int32_t iu);
-  int32_t RadtoIU(double rad);
+  /*
+   * Converts radians to encoder incremental units.
```
RutgerVanBeek:
```suggestion
   * Converts radians to encoder Incremental Units.
```

march_hardware/include/march_hardware/Encoder.h
```diff
@@ -58,23 +51,48 @@ class Encoder
...
+              << "lowerHardLimit: " << encoder.lower_limit_iu_ << ", "
+              << "upperSoftLimit: " << encoder.upper_soft_limit_iu_ << ", "
+              << "lowerSoftLimit: " << encoder.lower_soft_limit_iu_ << ", "
+              << "zeroPositionIU: " << encoder.zero_position_iu_;
   }
+
+  static const size_t MIN_RESOLUTION = 1;
+  static const size_t MAX_RESOLUTION = 32;
+
+  static constexpr double MAX_IU_PER_RAD_DIFFERENCE = 0.1;
```
RutgerVanBeek: Why a maximum? We already know what the IU per rad is for all the different joints. IU per rad should be constant and if it is not and something is going wrong somewhere.

RutgerVanBeek (commented):
Nice work on the tests. I think that automating this proces is really beneficial because of the automatic testing. I do not entirely agree on how you test whether the IUperRad is correct. First of all semantically is the IUperRad constant and is the range of motion the thing that is wrong. Furthermore from what I remember the linear joints have one bit resolution less, so defining a maximum for all joints doesn't seem a good idea. Lastly the range of motion can also be wrong in the other way that the IUperRad is too small. This is not tested.

march_hardware/include/march_hardware/Encoder.h
Olavhaasie: Actually, I think it should be `Internal Units`

march_hardware/include/march_hardware/Encoder.h
Olavhaasie: This maximum is used to give a warning when the IU per rad calculated from the encoder positions is 10% different than the IU per rad calculated from the given limits. So indeed something is wrong when the two values are different (which is happening right now, but that is another issue)

RutgerVanBeek (dismissed):

march_hardware/src/Encoder.cpp
```diff
@@ -1,122 +1,138 @@
...
+  if (this->isWithinSoftLimitsIU(target_iu))
   {
     return true;
   }
 
-  if (currentIU >= this->getUpperSoftLimitIU())
+  if (current_iu >= this->upper_soft_limit_iu_)
   {
-    return (targetIU <= currentIU) && (targetIU > this->getLowerSoftLimitIU());
+    return (target_iu <= current_iu) && (target_iu > this->lower_soft_limit_iu_);
```
Roelemans: This can be done more efficiently with the two (instead of three) conditions
`if (target_iu >= this->upper_soft_limit_iu_)
{
  return (target_iu <= current_iu)
}`

same for the statement below. You could then even remove the above if (this->isWithinSoftLimitsIU(target_iu))
  {
    return true;
  }
and simply return true if both of the if-statements suggested here return true.

Roelemans (changes requested):
Will look at TestEncoder.cpp later

RutgerVanBeek (dismissed):

march_hardware/src/Encoder.cpp
```diff
@@ -1,122 +1,133 @@
...
-#include <march_hardware/Encoder.h>
 #include <cmath>
+
 #include <ros/ros.h>
 
 namespace march
 {
-Encoder::Encoder(int numberOfBits, int32_t minPositionIU, int32_t maxPositionIU, int32_t zeroPositionIU,
-                 float safetyMarginRad)
+const double PI_2 = 6.28318530717958647692;
```
Joris-JW: could use 2*M_PI, this constant is defined in the cmath library

march_hardware/src/Encoder.cpp
```diff
@@ -1,122 +1,133 @@
...
+
+size_t Encoder::calculateTotalPositions(size_t number_of_bits)
+{
+  if (number_of_bits < Encoder::MIN_RESOLUTION || number_of_bits > Encoder::MAX_RESOLUTION)
+  {
+    throw error::HardwareException(error::ErrorType::INVALID_ENCODER_RESOLUTION,
+                                   "Encoder resolution of %d is not within range [%ld, %ld]", number_of_bits,
+                                   Encoder::MIN_RESOLUTION, Encoder::MAX_RESOLUTION);
+  }
+  return std::pow(2, number_of_bits) - 1;
```
Joris-JW: Why do - 1? The encoder can also have a value of 0. Would not make a lot of difference but just for the sake of clarity.

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+class TestEncoderParameterizedValidTarget : public TestEncoder,
+                                            public testing::WithParamInterface<std::tuple<int32_t, int32_t, bool>>
+{
+};
+
+TEST_F(TestEncoder, CorrectLimits)
+{
+  ASSERT_EQ(this->encoder.getLowerHardLimitIU(), this->lower_limit);
+  ASSERT_EQ(this->encoder.getUpperHardLimitIU(), this->upper_limit);
+}
```
Roelemans, line 52: Two asserts in one test?

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+
+TEST_F(TestEncoder, CorrectSoftLimits)
+{
+  ASSERT_EQ(this->encoder.getLowerSoftLimitIU(), 3144);
+  ASSERT_EQ(this->encoder.getUpperSoftLimitIU(), 44650);
```
Roelemans: I think you should give a calculation of these values so that people can see where they come from

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+}
+
+TEST(TestEncoderConstructor, ResolutionBelowRange)
+{
+  ASSERT_THROW(march::Encoder(0, 0, 0, 0, 0, 0, 0), march::error::HardwareException);
```
Roelemans: Bad test. This very same error is thrown if lower_limit == upper_limit. Hence, if you set the resolution to between 1 and 32 (which should be valid), the HardwareException is still thrown.

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+}
+
+TEST(TestEncoderConstructor, ResolutionAboveRange)
+{
+  ASSERT_THROW(march::Encoder(50, 0, 0, 0, 0, 0, 0), march::error::HardwareException);
```
Roelemans: Same here.

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+}
+
+TEST(TestEncoderConstructor, DifferentIuPerRad)
+{
+  ASSERT_THROW(march::Encoder(12, 2, 1, 1.0, 2.0, 1.0, 2.0), march::error::HardwareException);
```
Roelemans: Same here. The lower_limit_iu = 2 > upper_limit_iu = 1, which will error.
Maybe it is a good idea to make different custom exceptions for these errors?

march_hardware/test/TestEncoder.cpp
Roelemans: In these tests, it is a bit difficult to see which input is which variable. What you could do is define all inputs for a correct encoder in the constructor of the test suite, which you can reuse in every test. In these exceptions tests, you can use all these predefined parameters and redefine only the few of them that should trigger an error.

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
-
-class TestEncoder : public ::testing::Test
+class TestEncoder : public testing::Test
 {
 protected:
-  const float angle = 42;
+  const int32_t lower_limit = 2053;
+  const int32_t upper_limit = 45617;
+  march::Encoder encoder =
+      march::Encoder(17, lower_limit, upper_limit, -0.34906585, 1.745329252, -0.296705973, 1.692969374);
```
Roelemans: The lower_limit_rad and upper_limit_rad come from calculations using the lower_limit_iu and upper_limit_iu right? Maybe it is a good idea to explicitly do these calculations here so that people can see where the values come from.

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+{
+  ASSERT_THROW(march::Encoder(50, 0, 0, 0, 0, 0, 0), march::error::HardwareException);
+}
+
+TEST(TestEncoderConstructor, DifferentIuPerRad)
+{
+  ASSERT_THROW(march::Encoder(12, 2, 1, 1.0, 2.0, 1.0, 2.0), march::error::HardwareException);
+}
+
+TEST(TestEncoderConstructor, InvalidSoftLimits)
```
Roelemans:
```suggestion
TEST(TestEncoderConstructor, LowerSoftLimitAboveUpperSoftLimit)
```
#descriptivenaming

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+{
+  ASSERT_THROW(march::Encoder(17, 2053, 45617, -0.34906585, 1.745329252, 0.4, 0.1), march::error::HardwareException);
+}
+
+TEST(TestEncoderConstructor, InvalidLowerSoftLimit)
```
Roelemans:
```suggestion
TEST(TestEncoderConstructor, LowerSoftLimitBelowLowerHardLimit)
```
(or without the "Hard")

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+{
+  ASSERT_THROW(march::Encoder(0, 0, 0, 0, 0, 0, 0), march::error::HardwareException);
+}
+
+TEST(TestEncoderConstructor, ResolutionAboveRange)
+{
+  ASSERT_THROW(march::Encoder(50, 0, 0, 0, 0, 0, 0), march::error::HardwareException);
+}
+
+TEST(TestEncoderConstructor, DifferentIuPerRad)
```
Roelemans:
```suggestion
TEST(TestEncoderConstructor, IuPerRadNotMatchingResolution)
```
(not very sure about this one)

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+{
+  ASSERT_THROW(march::Encoder(17, 2053, 45617, -0.34906585, 1.745329252, 0.4, 0.1), march::error::HardwareException);
+}
+
+TEST(TestEncoderConstructor, InvalidLowerSoftLimit)
+{
+  ASSERT_THROW(march::Encoder(17, 2053, 45617, -0.34906585, 1.745329252, -0.4, 1), march::error::HardwareException);
+}
+
+TEST(TestEncoderConstructor, InvalidUpperSoftLimit)
```
Roelemans:
```suggestion
TEST(TestEncoderConstructor, UpperSoftLimitAboveUpperHardLimit)
```
(or without the "Hard")

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+{
+  ASSERT_THROW(march::Encoder(17, 2053, 45617, -0.34906585, 1.745329252, -0.3, 2.0), march::error::HardwareException);
+}
+
+INSTANTIATE_TEST_CASE_P(ParameterizedLimits, TestEncoderParameterizedLimits,
+                        testing::Values(std::make_tuple(2052, false), std::make_tuple(2053, false),
+                                        std::make_tuple(2054, true), std::make_tuple(45616, true),
+                                        std::make_tuple(45617, false), std::make_tuple(45618, false),
+                                        std::make_tuple(3000, true), std::make_tuple(0, false),
+                                        std::make_tuple(60001, false)));
```
Roelemans: This is some nice testing :). Maybe also a test for a negative value? (which would be in range if positive)

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+  ASSERT_THROW(march::Encoder(17, 2053, 45617, -0.34906585, 1.745329252, -0.4, 1), march::error::HardwareException);
+}
+
+TEST(TestEncoderConstructor, InvalidUpperSoftLimit)
+{
+  ASSERT_THROW(march::Encoder(17, 2053, 45617, -0.34906585, 1.745329252, -0.3, 2.0), march::error::HardwareException);
+}
+
+INSTANTIATE_TEST_CASE_P(ParameterizedLimits, TestEncoderParameterizedLimits,
+                        testing::Values(std::make_tuple(2052, false), std::make_tuple(2053, false),
```
Roelemans:
```suggestion
                        testing::Values(std::make_tuple(this->lower_limit-1, false), std::make_tuple(this->lower_limit, false),
```
and same for the things below

march_hardware/test/TestEncoder.cpp
```diff
@@ -1,29 +1,116 @@
...
+                                        std::make_tuple(3000, 3001, true), std::make_tuple(3144, 3144, true),
+                                        std::make_tuple(0, -1, false), std::make_tuple(0, 60000, false)));
+
+TEST_P(TestEncoderParameterizedValidTarget, IsValidTargetIU)
+{
+  const int32_t current_iu = std::get<0>(this->GetParam());
+  const int32_t target_iu = std::get<1>(this->GetParam());
+  const bool expected = std::get<2>(this->GetParam());
+  ASSERT_EQ(this->encoder.isValidTargetIU(current_iu, target_iu), expected);
 }
```
Roelemans, line 165: Maybe make tests  for calculateTotalPositions, fromRad and toRad?

Roelemans (changes requested):
Haha and here I though I would have few comments on this PR

march_hardware/test/TestEncoder.cpp
Olavhaasie: Unfortunately, I was not able to find a nice way to implement this, since the `INSTANTIATE_TEST_CASE_P` cannot access the `TestEncoder` static or member variables.

march_hardware/test/TestEncoder.cpp
Olavhaasie: `calculateTotalPositions` is private, so that is not possible, also it is implicitly tested through the constructor. I have added tests for `fromRad` and `toRad`

march_hardware/test/TestEncoder.cpp
Olavhaasie: I let the soft limits depend on the radian limits and a buffer

march_hardware/test/TestEncoder.cpp
```diff
@@ -96,6 +96,16 @@ TEST_F(TestEncoder, UpperSoftLimitHigherThanUpperHardLimit)
...
+TEST_F(TestEncoder, ZeroPositionRadToZeroPosition)
+{
+  ASSERT_EQ(this->encoder.fromRad(0.0), this->zero_position);
+}
+
+TEST_F(TestEncoder, ZeroPositionToZeroRadians)
+{
+  ASSERT_DOUBLE_EQ(this->encoder.toRad(this->zero_position), 0.0);
+}
+
```
Roelemans, line 115: Maybe also test something other than zero? Now maybe this will simply always return zero

march_hardware/test/TestEncoder.cpp
```diff
@@ -29,54 +41,67 @@ class TestEncoderParameterizedValidTarget : public TestEncoder,
...
+                              this->upper_limit_rad, 0.4, 0.1),
+               march::error::HardwareException);
 }
 
-TEST(TestEncoderConstructor, InvalidLowerSoftLimit)
+TEST_F(TestEncoder, LowerSoftLimitLowerThanLowerHardLimit)
 {
-  ASSERT_THROW(march::Encoder(17, 2053, 45617, -0.34906585, 1.745329252, -0.4, 1), march::error::HardwareException);
+  ASSERT_THROW(march::Encoder(this->resolution, this->lower_limit, this->upper_limit, this->lower_limit_rad,
+                              this->upper_limit_rad, -0.4, 1),
```
Roelemans:
```suggestion
                              this->upper_limit_rad, -0.4, this->upper_soft_limit_rad),
```

march_hardware/test/TestEncoder.cpp
```diff
@@ -29,54 +41,67 @@ class TestEncoderParameterizedValidTarget : public TestEncoder,
...
+               march::error::HardwareException);
 }
 
-TEST(TestEncoderConstructor, InvalidUpperSoftLimit)
+TEST_F(TestEncoder, UpperSoftLimitHigherThanUpperHardLimit)
 {
-  ASSERT_THROW(march::Encoder(17, 2053, 45617, -0.34906585, 1.745329252, -0.3, 2.0), march::error::HardwareException);
+  ASSERT_THROW(march::Encoder(this->resolution, this->lower_limit, this->upper_limit, this->lower_limit_rad,
+                              this->upper_limit_rad, -0.3, 2.0),
```
Roelemans:
```suggestion
                              this->upper_limit_rad, this->lower_soft_limit_rad, 2.0),
```

march_hardware/test/TestEncoder.cpp
```diff
@@ -29,54 +41,67 @@ class TestEncoderParameterizedValidTarget : public TestEncoder,
...
+               march::error::HardwareException);
 }
 
 INSTANTIATE_TEST_CASE_P(ParameterizedLimits, TestEncoderParameterizedLimits,
-                        testing::Values(std::make_tuple(2052, false), std::make_tuple(2053, false),
-                                        std::make_tuple(2054, true), std::make_tuple(45616, true),
-                                        std::make_tuple(45617, false), std::make_tuple(45618, false),
+                        testing::Values(std::make_tuple(2053 - 1, false), std::make_tuple(2053, false),
+                                        std::make_tuple(2053 + 1, true), std::make_tuple(45617 - 1, true),
```
Roelemans, line 125:
```suggestion
                                        std::make_tuple(this->lower_hard_limit + 1, true), std::make_tuple(this->upper_hard_limit - 1, true),
```
Also for line above and below

march_hardware/test/TestEncoder.cpp
```diff
@@ -86,11 +111,11 @@ TEST_P(TestEncoderParameterizedLimits, IsWithinHardLimits)
 }
 
 INSTANTIATE_TEST_CASE_P(ParameterizedSoftLimits, TestEncoderParameterizedSoftLimits,
-                        testing::Values(std::make_tuple(3143, false), std::make_tuple(3144, false),
-                                        std::make_tuple(3145, true), std::make_tuple(44649, true),
-                                        std::make_tuple(44650, false), std::make_tuple(44651, false),
+                        testing::Values(std::make_tuple(3095 - 1, false), std::make_tuple(3095, false),
+                                        std::make_tuple(3095 + 1, true), std::make_tuple(44699 - 1, true),
+                                        std::make_tuple(44699, false), std::make_tuple(44699 + 1, false),
```
Roelemans, line 140: Same here

Roelemans (changes requested):

march_hardware/test/TestEncoder.cpp
```diff
@@ -96,6 +96,16 @@ TEST_F(TestEncoder, UpperSoftLimitHigherThanUpperHardLimit)
...
 }
 
+TEST_F(TestEncoder, ZeroPositionRadToZeroPosition)
+{
+  ASSERT_EQ(this->encoder.fromRad(0.0), this->zero_position);
+}
+
+TEST_F(TestEncoder, ZeroPositionToZeroRadians)
+{
+  ASSERT_DOUBLE_EQ(this->encoder.toRad(this->zero_position), 0.0);
```
Roelemans, line 113: Also a nonzero test maybe :innocent: this will probably pass if you return nothing

march_hardware/test/TestEncoder.cpp
Olavhaasie: As I said earlier:
> Unfortunately, I was not able to find a nice way to implement this, since the INSTANTIATE_TEST_CASE_P cannot access the TestEncoder static or member variables.

This is not possible and I was not able to find a way to do this :cry:

Roelemans: Sorry, missed that comment. I will deal with this then :+1:

Roelemans (approved):
Good job doing math haha :)

Joris-JW (approved):
nice

PR #219, opened 2020-02-22
Remove reset imc

Olavhaasie:
## Description
The new PCBs eliminated the encoder reset problem and therefore the reset IMotionCubes is no longer necessary. This will cause the joints to startup significantly faster :rocket: 

## Changes
* Remove `resetIMotionCube`


Roelemans:
Maybe good to test this during the training first? Or we can just revert I guess

Roelemans (approved):

RutgerVanBeek (approved):
Let's try this, but the fact that we haven't had encoder resets last week, does not mean they are gone for sure.

PR #220, opened 2020-02-26
Feature: default to effort mode

Olavhaasie:
Closes PM-287

## Description
Sets default controllers to effort and add watchdog timer. The PID tuning is default to 0 for now to avoid wiggles during startup. You have to launch the `march_gain_scheduling_node` in order to set the PIDs.

## Changes
* Set default actuation mode for IMCs to torque 
* Set default ros controller to effort
* Change naming in `EthercatMaster`
* Change control loop to 250 Hz
* Set PID tuning to 0


march_hardware_interface/config/march4/controllers.yaml
```diff
@@ -22,15 +22,15 @@ march:
...
-#        right_hip_aa: {p: 290, i: 0, d: 5, i_clamp: 100, publish_state: true, antiwindup: true} # TODO(BaCo) tune HAA joints
-#        right_hip_fe: {p: 80, i: 0, d: 10, i_clamp: 100, publish_state: true, antiwindup: true}
-#        right_knee: {p: 50, i: 0, d: 10, i_clamp: 100, publish_state: true, antiwindup: true}
+      gains: # Required because we're controlling an effort interface
+        left_ankle: {p: 0, i: 0, d: 0, i_clamp: 100, publish_state: true, antiwindup: true}
+        left_hip_aa: {p: 0, i: 0, d: 0, i_clamp: 100, publish_state: true, antiwindup: true}
+        left_hip_fe: {p: 0, i: 0, d: 0, i_clamp: 100, publish_state: true, antiwindup: true}
+        left_knee: {p: 0, i: 0, d: 0, i_clamp: 100, publish_state: true, antiwindup: true}
+        right_ankle: {p: 0, i: 0, d: 0, i_clamp: 100, publish_state: true, antiwindup: true}
+        right_hip_aa: {p: 0, i: 0, d: 0, i_clamp: 100, publish_state: true, antiwindup: true} # TODO(BaCo) tune HAA joints
```
Roelemans:
```suggestion
        right_hip_aa: {p: 0, i: 0, d: 0, i_clamp: 100, publish_state: true, antiwindup: true}
```

march_hardware_interface/src/march_hardware_interface_node.cpp
```diff
@@ -41,7 +41,7 @@ int main(int argc, char** argv)
     return 1;
   }
 
-  const double loop_hz = ros::param::param("~loop_hz", 100.0);
+  const double loop_hz = ros::param::param("~loop_hz", 250.0);
```
Roelemans: Can you link this directly to the ethercat cycle time so that we don't have to change both when increasing the frequency?

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -43,41 +46,54 @@ void EthercatMaster::start(std::vector<Joint>& joints)
...
   }
   ROS_INFO("%d slave(s) found and initialized.", slave_count);
 }
 
+int setSlaveWatchdogTimer(uint16 slave)
+{
+  uint16 configadr = ec_slave[slave].configadr;
+  ec_FPWRw(configadr, 0x0400, IMotionCube::WATCHDOG_DIVIDER, EC_TIMEOUTRET);  // Set the divider register of the WD
+  ec_FPWRw(configadr, 0x0410, IMotionCube::WATCHDOG_TIME, EC_TIMEOUTRET);     // Set the PDI watchdog = WD
+  ec_FPWRw(configadr, 0x0420, IMotionCube::WATCHDOG_TIME, EC_TIMEOUTRET);     // Set the SM watchdog = WD
```
Roelemans, line 77: There are so many things I don't understand. Was this wisdom bestowed to us by our great IMC god Vlad, or is this somewhere in the manual? @Deakensbubbles

Roelemans (changes requested):

march_hardware_interface/config/march4/controllers.yaml
RutgerVanBeek:
```suggestion
        right_hip_aa: {p: 0, i: 0, d: 0, i_clamp: 100, publish_state: true, antiwindup: true} 
```
Please remove BaCo todo

march_hardware_interface/src/march_hardware_interface_node.cpp
RutgerVanBeek: Is this increased to the frequency of ethercat, because of the watchdog, because then this should forced to be equal by using the same parameter from the yaml?

Olavhaasie: Good idea, could it be useful to add a parameter that specifies a multiplier of the frequency, so that we can set the controller frequency lower, but still a multiple of the ethercat cycle?

Olavhaasie: Yes it was increased to synchronize the controller and ethercat loop, see my other comment for deriving this from the ethercat cycle time.

RutgerVanBeek: Yes, I think when you read the ecatcycletime in the builder that we can make a parameter out of it. We should make the frequency here dependent on that parameter.

march_hardware/src/EtherCAT/EthercatMaster.cpp
Joris-JW: Should we not move the SDO messages to the writeInitialSettings in the imotioncube cpp?

Olavhaasie: These are not SDO messages and I am not sure if you move this there that this still works. We found that we had to explicitly use the `PO2SOconfig` callback. Perhaps @Deakensbubbles can tell you more about this?

Joris-JW: that would be nice.

Joris-JW (commented):
some additional information about the watchdog configurations would be nice.

Roelemans (approved):

RutgerVanBeek (approved):
Nice work with parameters

march_hardware/src/EtherCAT/EthercatMaster.cpp
RPKrijnen: These are indeed not actual SDO messages, but make use of the read and write functions during a so called hook function. This knowledge was bestowed upon us by our two gods of wisdom; Vlad - the knower of all IMC and Arthur Ketels - The Wisest wiseman of wisdom. 
Jokes aside; Vlad let us know about the registers (0x0400 etc.) and their meaning and Arthur helped us with changing the values of these registers. These registers are somewhat hidden away, either by Technosoft or Beckhoff and therefore require a more difficult way of changing these values.

Joris-JW: Is there any way to get more of these hidden registers? maybe some kind of general ethercat protocol that AK knows about? maybe there is some more useful stuff. Either way, this is nice!

Joris-JW (approved):

PR #221, opened 2020-02-27
Upload URDF instead of xacro

Olavhaasie:
## Description
Removed the uploading of the xacro and replaced it by uploading the URDF, which is built by `march_description` since project-march/march#457

_(You also no longer get the warning which says that in order processing is the default since lunar. This caused my tab completion on `roslaunch` to do weird stuff.)_


Joris-JW (approved):

Roelemans (approved):
What do you mean with the tab after roslaunch comment? Is that problem fixed by this PR?

PR #222, opened 2020-02-28
Revert "Feature: add zero position calculation (#217)"

Roelemans:
This reverts commit 51997bbe

When doing a home sit, the knees and HAAs (and possibly the other joints as well) shot towards their endstops. This commit should cause that problem.

A strange point about this problem was that there was a delay after homesit before this problem appeared. It seemed to only happen after the gait was finished, but I am not certain, we should check the logs.


RutgerVanBeek (commented):
Is it tested, whether this makes the difference? May be do that first.

Olavhaasie:
This is not an issue, the gain scheduling has some bugs

PR #223, opened 2020-02-28
Fix oopsie by using seconds instead of milliseconds

Olavhaasie:
##  Description
Using seconds instead of milliseconds is not a good idea for a control loop :crying_cat_face:


RutgerVanBeek (approved):

Roelemans (approved):
Knap hoor

Ciliaclaij (approved):

Roelemans (approved):

PR #224, opened 2020-02-28
Fix forgeting to add ros::Duration

Olavhaasie:
## Description
I did an oopsie on the last oopsie, but now It is fixed for good :v:


Roelemans (approved):

RutgerVanBeek (approved):
Oopsie oopsie

PR #225, opened 2020-03-04
Feature: add incremental encoder

Olavhaasie:
Closes PM-24

## Description
Adds the feature to read out the incremental encoder on the joints. It maps the Motor Position object from the IMotionCube.

@bjornminderman started on this feature and I solved the conflicts it had, since it was a bit outdated.

The original `Encoder` class is now a super class of `AbsoluteEncoder` and `IncrementalEncoder`. Also in order to keep it clean I have added a new `encoder` directory, where these classes reside. The `IMotionCube` and `Joint` class now have two separate functions for reading out the absolute and incremental encoder position. The incremental encoder position is only published on the imc state topic. It shouldn't be too hard to use it for any calculations in the hardware interface.

## Changes
* Create `AbsoluteEncoder` and `IncrementalEncoder`
* Rename a lot of stuff from `Encoder` to `AbsoluteEncoder`
* Add publishing incremental encoder position
* Add tests


march_hardware/include/march_hardware/encoder/Encoder.h
```diff
@@ -0,0 +1,48 @@
...
+
+namespace march
+{
+class Encoder
+{
+public:
+  explicit Encoder(size_t number_of_bits);
+
+  /*
+   * Reads out the encoder from the slave and returns the value in Internal Units (IU).
```
RutgerVanBeek, line 17: Absolute or incremental encoder? or change the encoder to an encoder. Now it reads like there is only one encoder.

march_hardware/src/encoder/IncrementalEncoder.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+double IncrementalEncoder::getAngleRad(uint8_t byte_offset)
+{
+  return this->toRad(Encoder::getAngleIU(byte_offset));
+}
+
+double IncrementalEncoder::toRad(int32_t iu)
+{
+  return iu * 2 * M_PI / Encoder::getTotalPositions();
+}
```
RutgerVanBeek: Although I am not sure how it does exactly work, I think you cannot do this as it can wrap around. I think we should leave this or a similar function out until we know what the output is. Or does @bjornminderman know more about this?

RutgerVanBeek (changes requested):
Nice work @bjornminderman and @Olavhaasie! Main comment is that I think we should for know just read the integer value. The current conversion to radians does not seem appropriate, because we use the total positions, but for an incremental encoder this is bigger than one round.

march_hardware/src/encoder/IncrementalEncoder.cpp
```diff
@@ -0,0 +1,21 @@
...
+namespace march
+{
+IncrementalEncoder::IncrementalEncoder(size_t number_of_bits) : Encoder(number_of_bits)
+{
+}
+
+double IncrementalEncoder::getAngleRad(uint8_t byte_offset)
+{
+  return this->toRad(Encoder::getAngleIU(byte_offset));
+}
```
Roelemans, line 9: You can just move this to the superclass since AbsoluteEncoder contains an identical method

march_hardware/src/encoder/IncrementalEncoder.cpp
Roelemans: I think you can do the same here if you set zero_position_iu to zero.

march_hardware/include/march_hardware/encoder/Encoder.h
Olavhaasie: It's both actually since this is a base class for those types of encoders. The _the_ refers to the actual encoder instance in use. So this description would be valid for both `absolute_encoder.getAngleIU()` and `incremental_encoder.getAngleIU()`

march_hardware/src/encoder/IncrementalEncoder.cpp
Olavhaasie: I agree, I am not sure if it will be useful, for now the IU is published and the radians are not used.

Roelemans (changes requested):

Olavhaasie:
I will study the behavior of the incremental encoders first

Olavhaasie:
I have added a `transmission` variable to the incremental encoder config in https://github.com/project-march/hardware-interface/pull/225/commits/b53791052fb7ff869baa17dbf90d4f691b60dc13. The transmission is already defined in the xacro, however, I was not able to access that with the `urdf` package, so I added it as a variable in the yaml.

march_hardware/src/encoder/IncrementalEncoder.cpp
```diff
@@ -16,6 +17,6 @@ double IncrementalEncoder::getAngleRad(uint8_t byte_offset)
```
Roelemans: Move above method to super class? :innocent:

march_hardware/src/encoder/IncrementalEncoder.cpp
```diff
@@ -16,6 +17,6 @@ double IncrementalEncoder::getAngleRad(uint8_t byte_offset)
 
 double IncrementalEncoder::toRad(int32_t iu)
 {
-  return iu * 2 * M_PI / Encoder::getTotalPositions();
+  return iu * this->transmission_ * 2 * M_PI / Encoder::getTotalPositions();
```
Roelemans:
```suggestion
  return iu * this->transmission_ * PI_2 / Encoder::getTotalPositions();
```

march_hardware/test/encoder/TestIncrementalEncoder.cpp
```diff
@@ -20,6 +21,6 @@ TEST_F(TestIncrementalEncoder, ZeroIUToRad)
 TEST_F(TestIncrementalEncoder, CorrectToRad)
 {
   const int32_t iu = 1000;
-  const double expected = iu * 2.0 * M_PI / std::pow(2, this->resolution);
+  const double expected = iu * this->transmission * 2.0 * M_PI / std::pow(2, this->resolution);
```
Roelemans, line 24:
```suggestion
  const double expected = iu * this->transmission * PI_2 / std::pow(2, this->resolution);
```

Roelemans (changes requested):

RutgerVanBeek (dismissed):

Roelemans (approved):
:)

RutgerVanBeek (approved):

PR #226, opened 2020-03-04
Replace obscure smoothing method normal logic

Roelemans:
Replace obscure smoothing method with simple math that does exactly the same.


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -210,7 +210,7 @@ void MarchHardwareInterface::read(const ros::Time& /* time */, const ros::Durati
 
     // Apply exponential smoothing to velocity obtained from encoder with
     // alpha=0.2
-    joint_velocity_[i] = filters::exponentialSmoothing(joint_velocity, joint_velocity_[i], 0.2);
+    joint_velocity_[i] = 0.2 * joint_velocity + 0.8 * joint_velocity_[i];
```
Olavhaasie: It think it's better to use a constant here declared in `MarchHardwareInterface` something like
```c++
public:
...
static constexpr double ALPHA = 0.2;
...
```
Then change this line to
```suggestion
    joint_velocity_[i] = MarchHardwareInterface::ALPHA * joint_velocity + (1 - MarchHardwareInterface:ALPHA) * joint_velocity_[i];
```

Olavhaasie (changes requested):
You should also remove the `control_toolbox` dependencies in https://github.com/project-march/hardware-interface/blob/07046a4ae05ecc41d8cd65fe57ab2f0f40d3aac2/march_hardware_interface/CMakeLists.txt#L7 and https://github.com/project-march/hardware-interface/blob/07046a4ae05ecc41d8cd65fe57ab2f0f40d3aac2/march_hardware_interface/package.xml#L20

If we ever want to use them again, we can simply revert them, better to remove it to avoid cluttering the dependencies if we forget to remove it later.

Olavhaasie (approved):

RutgerVanBeek (approved):

PR #229, opened 2020-03-06
Fix: remove debug statements in real time safe loop

Olavhaasie:
## Description
Removed debug statements in real time safe loop, to avoid nondeterministic behavior when printing.

See [this thread](https://answers.ros.org/question/302486/does-ros_info-break-realtime/) on ROS answers.

## Changes
* Remove `ROS_DEBUG` statements
* Remove redundant `ROS_ERROR` and `ROS_FATAL` before throwing exception

### DIFF
--- a/march_hardware_interface/src/march_hardware_interface.cpp
+++ b/march_hardware_interface/src/march_hardware_interface.cpp
@@ -213,8 +213,6 @@ void MarchHardwareInterface::read(const ros::Time& /* time */, const ros::Durati
         MarchHardwareInterface::ALPHA * joint_velocity + (1 - MarchHardwareInterface::ALPHA) * joint_velocity_[i];
 
     joint_effort_[i] = joint.getTorque();
-
-    ROS_DEBUG("Joint %s: read position %f", joint_names_[i].c_str(), joint_position_[i]);
   }
 
   this->updateIMotionCubeState();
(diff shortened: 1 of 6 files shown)


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -213,20 +213,13 @@ void MarchHardwareInterface::read(const ros::Time& /* time */, const ros::Durati
...
 
   if (has_power_distribution_board_)
   {
     power_distribution_board_read_ = march_robot_.getPowerDistributionBoard();
-
-    if (!power_distribution_board_read_.getHighVoltage().getHighVoltageEnabled())
-    {
-      ROS_WARN_THROTTLE(10, "All-High-Voltage disabled");
-    }
   }
```
RutgerVanBeek, line 223: I think this is a very useful statement if we want to run with powerdistributionboard again

RutgerVanBeek (changes requested):

march_hardware_interface/src/march_hardware_interface.cpp
Olavhaasie: True, however it also printed here https://github.com/project-march/hardware-interface/blob/c5fdaffece8cf0b8b137124faec1034743c7bcce/march_hardware_interface/src/march_hardware_interface.cpp#L327-L330

I found the above location a better one

Roelemans (approved):

RutgerVanBeek (approved):

PR #232, opened 2020-03-06
Fix: ethercat loop

Olavhaasie:
## Description
Fixed an issue in `EthercatMaster` where the `is_operational_` member could be written to and read from at the same time. This results in undefined behaviour (even though it is a bool). So I changed it to a `std::atomic<bool>` which guarantees atomic reading and writing.

Furthermore, the `ethercatLoop` was using `boost::chrono` for its timing. However, Boost was not included in the project and the same functionality exists in the standard library. I changed it to using durations, which is a lot nicer to read, since you no longer have to convert microseconds to milliseconds and so on.

## Changes
* Changed `is_operational_` type from `bool` to `std::atomic<bool>`
* Use `std::chrono` instead of `boost::chrono`
* Add some `this->` to members to be more consistent and distinct them from the `soem` library functions

### DIFF
--- a/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
+++ b/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
@@ -1,6 +1,7 @@
 // Copyright 2019 Project March.
 #ifndef MARCH_HARDWARE_ETHERCAT_ETHERCATMASTER_H
 #define MARCH_HARDWARE_ETHERCAT_ETHERCATMASTER_H
+#include <atomic>
 #include <vector>
 #include <string>
 #include <thread>
@@ -28,7 +29,7 @@ public:
   EthercatMaster& operator=(const EthercatMaster&) = delete;
 
   /* Enable the move constructor */
-  EthercatMaster(EthercatMaster&&) = default;
+  EthercatMaster(EthercatMaster&&) = delete;
   EthercatMaster& operator=(EthercatMaster&&) = delete;
 
   bool isOperational() const;
@@ -77,7 +78,7 @@ private:
    */
   static void monitorSlaveConnection();
 
-  bool is_operational_ = false;
+  std::atomic<bool> is_operational_;
 
   const std::string ifname_;
   const int max_slave_index_;
--- a/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
+++ b/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
@@ -28,7 +28,7 @@ public:
   EthercatMaster(const EthercatMaster&) = delete;
   EthercatMaster& operator=(const EthercatMaster&) = delete;
 
-  /* Enable the move constructor */
+  /* Delete move constructor/assignment since atomic bool can not be moved */
   EthercatMaster(EthercatMaster&&) = delete;
   EthercatMaster& operator=(EthercatMaster&&) = delete;
 
--- a/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
+++ b/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
@@ -71,7 +71,7 @@ private:
   /**
    * Sends the PDO and receives the working counter and check if this is lower than expected.
    */
-  void SendReceivePDO();
+  void sendReceivePdo();
 
   /**
    * Checks if all the slaves are connected and in operational state.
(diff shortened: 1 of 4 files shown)


march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
```diff
@@ -27,8 +28,8 @@ class EthercatMaster
   EthercatMaster(const EthercatMaster&) = delete;
   EthercatMaster& operator=(const EthercatMaster&) = delete;
 
-  /* Enable the move constructor */
-  EthercatMaster(EthercatMaster&&) = default;
+  /* Delete move constructor/assignment since atomic bool can not be moved */
```
Roelemans:
```suggestion
  /* Delete move constructor/assignment since atomic bool cannot be moved */
```
what is a move constructor? :sweat_smile:

march_hardware/include/march_hardware/MarchRobot.h
```diff
@@ -37,7 +37,7 @@ class MarchRobot
   MarchRobot& operator=(const MarchRobot&) = delete;
 
   /* Enable the move constructor */
```
Roelemans:
```suggestion
  /* Delete move constructor/assignment since EthercatMaster cannot be moved */
```

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -142,45 +141,46 @@ void EthercatMaster::ethercatSlaveInitiation(std::vector<Joint>& joints)
...
-  size_t rate = 1000 / cycle_time_ms_;
+  size_t total_loops = 0;
+  size_t not_achieved_count = 0;
+  const size_t rate = 1000 / this->cycle_time_ms_;
+  const std::chrono::milliseconds cycle_time(this->cycle_time_ms_);
 
   while (this->is_operational_)
   {
-    auto start = boost::chrono::high_resolution_clock::now();
+    const auto begin = std::chrono::high_resolution_clock::now();
```
Roelemans:
```suggestion
    const auto start_time = std::chrono::high_resolution_clock::now();
```

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -142,45 +141,46 @@ void EthercatMaster::ethercatSlaveInitiation(std::vector<Joint>& joints)
...
 
-    SendReceivePDO();
+    this->sendReceivePdo();
     monitorSlaveConnection();
 
-    auto stop = boost::chrono::high_resolution_clock::now();
-    auto duration = boost::chrono::duration_cast<boost::chrono::microseconds>(stop - start);
+    const auto end = std::chrono::high_resolution_clock::now();
```
Roelemans:
```suggestion
    const auto end_time = std::chrono::high_resolution_clock::now();
```

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -142,45 +141,46 @@ void EthercatMaster::ethercatSlaveInitiation(std::vector<Joint>& joints)
...
+    const auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end - begin);
```
Roelemans:
```suggestion
    const auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end_time - begin_time);
```

Roelemans (changes requested):

Roelemans:
Maybe we should test this on exo first? An oopsie on cycle times can be a bit annoying as you probably know :smile_cat:

march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
Olavhaasie: A move constructor is a constructor which accepts a reference to an instance of its type, then copies all values from that instance and leaves the instance with default values.
https://en.cppreference.com/w/cpp/language/move_constructor

Olavhaasie:
@Roelemans I have reverted https://github.com/project-march/hardware-interface/pull/232/commits/41b5ec389ab78093f9150a67d8d4d1e46a5b3c1d please create a separate PR for this.

Joris-JW (approved):
nice (but indeed test on exo)

Roelemans (approved):

PR #233, opened 2020-03-06
Fix/stop ethercat when imc fault

Roelemans:
## Description

When an IMotionCube goes to fault state, it takes approximately 0.4 seconds before the master shuts ethercat down. If this happens due to something that is not an ethercat error, there will be seven IMC that do not go to fault state until the master shuts down ethercat. However, the master no longer updates the efforts, so the IMCs continue sending their previous effort as long as ethercat is active. Hence it takes 0.45 seconds (0.4 + watchdog) until the joints do a quickstop. 

This PR stops ethercat immediately after a fault state is detected, instead of waiting for the destructor of MarchRobot to start.


bjornminderman (approved):
Good fix!

RPKrijnen (approved):
Nicely found!

PR #234, opened 2020-03-09
Feature/pm 308 velocity using incremental encoder

Roelemans:
Closes PM-308

## Description

Uses the incremental encoder to determine velocity for the linear joints. This has a much higher resolution allowing less smoothing and a higher accuracy.

### DIFF
--- a/march_hardware/include/march_hardware/IMotionCube.h
+++ b/march_hardware/include/march_hardware/IMotionCube.h
@@ -28,6 +28,7 @@ public:
 
   double getAngleRadAbsolute();
   double getAngleRadIncremental();
+  double getAngleRadMostPrecise();
   int16_t getTorque();
   int32_t getAngleIUAbsolute();
   int32_t getAngleIUIncremental();
@@ -84,6 +85,7 @@ private:
   AbsoluteEncoder absolute_encoder_;
   IncrementalEncoder incremental_encoder_;
   ActuationMode actuation_mode_;
+  bool incremental_more_precise_;
 
   std::unordered_map<IMCObjectName, uint8_t> miso_byte_offsets_;
   std::unordered_map<IMCObjectName, uint8_t> mosi_byte_offsets_;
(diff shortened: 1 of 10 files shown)


march_hardware/include/march_hardware/IMotionCube.h
```diff
@@ -83,6 +84,7 @@ class IMotionCube : public Slave
   AbsoluteEncoder absolute_encoder_;
   IncrementalEncoder incremental_encoder_;
   ActuationMode actuation_mode_;
+  bool incremental_more_precise_;
```
Joris-JW:
```suggestion
  bool is_incremental_more_precise_;
```

Joris-JW (commented):
Have u looked into a way of removing the noise of the incremental encoder data line of the rotational joint. If this is indeed more precise it maybe useful? We are probably not the first!

RPKrijnen (commented):
I fear that over time the incremental encoder might become less precise than the absolute encoder. This would be caused by backlash in the motor (however slight). It could be useful to continuously compare the value of the incremental encoder with the absolute encoder. 
The absolute encoder does not experience backlash and is therefore capable of providing a reference signal. Frequent enough updates can cause the drift of the signal to minimize while still maintaining high precision.

RutgerVanBeek:
> I fear that over time the incremental encoder might become less precise than the absolute encoder. This would be caused by backlash in the motor (however slight). It could be useful to continuously compare the value of the incremental encoder with the absolute encoder.
> The absolute encoder does not experience backlash and is therefore capable of providing a reference signal. Frequent enough updates can cause the drift of the signal to minimize while still maintaining high precision.

You are right, therefore we decided to start of easy and use the incremental encoder for velocity calculations only. Then we should not have the drifting problem. In the future we might also choose to use the incremental encoder for the position, then indeed frequent updates seem necessary.

Roelemans:
> Have u looked into a way of removing the noise of the incremental encoder data line of the rotational joint. If this is indeed more precise it maybe useful? We are probably not the first!

Certainly. You can probably combine the incremental encoder with the absolute for a much better signal. This will require some research though.

Joris-JW (dismissed):
yeeeet

Olavhaasie (approved):

RutgerVanBeek (approved):

PR #236, opened 2020-03-11
max effort change

RutgerVanBeek:
Closes PM-295

## Description

This is not tested on exo, but feedback on the implementation is already welcome. The effort sometimes goes up a lot and goes down right after or the other way around. This for instance usually happen with the effort peaks during a short circuit. Next to that it is safer to not let the effort go to max effort at once. The safety limits always override this feature. 

I think the effect on the performance will be little, because the effort can still change from min to max in 32 ms and the effort peaks usually consist of just one point. It will reduce the effect of mistakes in the controller and I think it will reduce the probability of short-circuits. However the effect of this definitely needs to be tested. 
## Changes

* Added a variable that stores a copy of the joint_effort_command
* Limit the change in effort with the MAX_EFFORT_CHANGE, which is set to 5000


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -230,6 +231,12 @@ void MarchHardwareInterface::write(const ros::Time& /* time */, const ros::Durat
   {
     // Enlarge joint_effort_command because ROS control limits the pid values to a certain maximum
     joint_effort_command_[i] = joint_effort_command_[i] * 1000.0;
+    if (abs(joint_effort_command_copy_[i] - joint_effort_command_[i]) > MAX_EFFORT_CHANGE)
```
Olavhaasie: Prefix standard library functions with `std::` so that it is clear they are from the standard library (and not from us)
```suggestion
    if (std::abs(joint_effort_command_copy_[i] - joint_effort_command_[i]) > MAX_EFFORT_CHANGE)
```

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -230,6 +231,12 @@ void MarchHardwareInterface::write(const ros::Time& /* time */, const ros::Durat
...
+    {
+      joint_effort_command_copy_[i] =
+          joint_effort_command_copy_[i] +
+          copysign(MAX_EFFORT_CHANGE, joint_effort_command_[i] - joint_effort_command_copy_[i]);
```
Olavhaasie: Same here
```suggestion
          std::copysign(MAX_EFFORT_CHANGE, joint_effort_command_[i] - joint_effort_command_copy_[i]);
```

march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
```diff
@@ -113,6 +114,7 @@ class MarchHardwareInterface : public hardware_interface::RobotHW
 
   std::vector<double> joint_effort_;
   std::vector<double> joint_effort_command_;
+  std::vector<double> joint_effort_command_copy_;
```
Olavhaasie: I would call this vector `joint_last_effort_command_` since that better describes what it contains.

march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
```diff
@@ -84,6 +84,7 @@ class MarchHardwareInterface : public hardware_interface::RobotHW
 
   /* Exponential smoothing constant of the velocity */
   static constexpr double ALPHA = 0.2;
+  static constexpr double MAX_EFFORT_CHANGE = 5000;
```
Olavhaasie, line 88: Perhaps add a comment to describe what this constant means? Without having to read the code where it is used

march_hardware_interface/src/march_hardware_interface.cpp
Olavhaasie: You write to the `joint_effort_command_copy_` here, however, you never read this value again. Instead you write over it with the `joint_effort_command` on line 254. I think you should write to `joint_effort_command_` here.

Olavhaasie (changes requested):

march_hardware_interface/src/march_hardware_interface.cpp
Roelemans: Instead of using this if and copysign structure, you could also use 
`joint_effort_command_[i] = std::min(joint_effort_command_copy_[i] + MAX_EFFORT_CHANGE, std::max(joint_effort_command_copy_[i] - MAX_EFFORT_CHANGE), joint_effort_command_[i])`

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -2,6 +2,7 @@
 #include "march_hardware_interface/march_hardware_interface.h"
 #include "march_hardware_interface/power_net_on_off_command.h"
 
+#include <cmath>
```
Roelemans, line 5: What do you use this for?

Olavhaasie: `std::abs` and `std::copysign`

march_hardware_interface/src/march_hardware_interface.cpp
RutgerVanBeek: Why would this be better? I think this follows the intuition better, where you take the old value + the max effort in the direction of the change.

Olavhaasie (dismissed):

march_hardware_interface/src/march_hardware_interface.cpp
Roelemans: Was intended as a suggestion. Use whatever has your preference :+1:

Roelemans (dismissed):

Olavhaasie (approved):

Roelemans (approved):

Joris-JW (approved):

PR #237, opened 2020-03-13
Feature: custom exceptions for Robot, Joint and Slave

Olavhaasie:
Closes PM-318 PM-319 PM-320

## Description
This PR adds custom exceptions for the `MarchRobot`, `Joint` and `Slave` classes. I started on the first, but then I figured I could just do them all in one go, since otherwise I would get conflicts on the error types and codes.

I have also added tests for the new exceptions and added a whole new test case for the `Slave` class. This class was implicitly tested through `IMotionCubeTest` and `TemperatureGESTest`, so I removed them there and created a new test.

## Changes
* New error types
* New tests


march_hardware/src/Joint.cpp
```diff
@@ -23,30 +24,24 @@ void Joint::initialize(int ecatCycleTime)
...
-    ROS_INFO("[%s] Preparing for actuation", this->name.c_str());
-    this->iMotionCube.goToOperationEnabled();
-    ROS_INFO("[%s] Successfully prepared for actuation", this->name.c_str());
-  }
-  else
-  {
-    ROS_ERROR("[%s] Trying to prepare for actuation while it is not "
-              "allowed to actuate",
-              this->name.c_str());
+    throw error::HardwareException(error::ErrorType::NOT_ALLOWED_TO_ACTUATE, "Trying to prepare joint %s for actuation",
```
Joris-JW: The message suggests that the program is still trying to prepare the joint for actuation, which is not the case.
```suggestion
    throw error::HardwareException(error::ErrorType::NOT_ALLOWED_TO_ACTUATE, "Failed to prepare joint %s for actuation",
```

Joris-JW (changes requested):

Joris-JW (dismissed):

march_hardware/src/Joint.cpp
```diff
@@ -23,30 +24,24 @@ void Joint::initialize(int ecatCycleTime)
...
-                 "yet its actuate method has been called",
-                 this->name.c_str());
-  // TODO(BaCo) check that the position is allowed and does not exceed (torque)
-  // limits.
+  if (!this->allowActuation)
+  {
+    throw error::HardwareException(error::ErrorType::NOT_ALLOWED_TO_ACTUATE, "Joint %s is not allowed to actuate",
+                                   this->name.c_str());
+  }
+  // TODO(BaCo) check that the position is allowed and does not exceed (torque) limits.
```
Roelemans: Should we make an issue for this?

Roelemans (dismissed):

march_hardware/src/Joint.cpp
Olavhaasie: Maybe we should check all the `TODO`s in the code and create issues from them if necessary. I'll do it right now

Joris-JW (approved):

Roelemans (approved):

PR #240, opened 2020-03-13
Feature: filter outliers from encoders

Roelemans:
Closes PM-327

## Description

The encoder data sometimes has strange outliers that are clearly incorrect. We can filter them out by comparing the output from the two encoders. When the motor displacement they measure is larger than two times their resolution, we can assume that one of them has measured an outlier. In that case, use whichever encoder measures a speed that is closest to the speed from the last timestep.

## Changes

* Combine both encoder inputs to filter outliers
* Move calculations of velocity and position to the Joint class so the HardwareInterface does not need to know that there are two encoders
* Use incremental encoder to determine position, but re-calibrate whenever it drifts too far from the absolute encoder value
* Remove smoothing of velocity


march_hardware/test/encoder/TestIncrementalEncoder.cpp
```diff
@@ -22,5 +22,5 @@ TEST_F(TestIncrementalEncoder, CorrectToRad)
 {
   const int32_t iu = 1000;
   const double expected = iu * 2.0 * M_PI / (std::pow(2, this->resolution) * this->transmission);
-  ASSERT_EQ(expected, this->encoder.toRad(iu));
+  ASSERT_TRUE(std::abs(expected - this->encoder.toRad(iu)) < std::pow(10, -9));
```
RutgerVanBeek: Is there anassertAlmostEqual in c++?

march_hardware/src/Joint.cpp
```diff
@@ -50,34 +55,62 @@ void Joint::actuateRad(double targetPositionRad)
...
+    }
   }
-  return this->iMotionCube.getAngleRadAbsolute();
+  else
+  {
+    // Recalibrate joint position if it has drifted
+    if (std::abs(this->absolute_position_ - this->position_) > 2 * this->iMotionCube.getAbsoluteRadPerBit())
+    {
+      this->position_ = this->absolute_position_;
+      ROS_INFO("Recalibrated joint position %s", this->name.c_str());
```
RutgerVanBeek: Good for testing purposes, of course not for the real implementation, because I think drift in incremental encoders will happen often.

march_hardware/src/IMotionCube.cpp
```diff
@@ -22,9 +22,6 @@ IMotionCube::IMotionCube(int slave_index, AbsoluteEncoder absolute_encoder, Incr
 {
   this->absolute_encoder_.setSlaveIndex(slave_index);
   this->incremental_encoder_.setSlaveIndex(slave_index);
-  this->is_incremental_more_precise_ =
-      (this->incremental_encoder_.getTotalPositions() * this->incremental_encoder_.getTransmission() >
-       this->absolute_encoder_.getTotalPositions() * 10);
   // Multiply by ten to ensure the rotational joints keep using absolute encoders. These are somehow more accurate
   // even though they theoretically shouldn't be.
```
RutgerVanBeek: Also remove the corresponding comment

march_hardware/src/Joint.cpp
```diff
@@ -50,34 +55,62 @@ void Joint::actuateRad(double targetPositionRad)
...
 {
-  if (!hasIMotionCube())
+  const double new_incremental_position = this->iMotionCube.getAngleRadIncremental();
+  const double new_absolute_position = this->iMotionCube.getAngleRadAbsolute();
+
+  // Get velocity from encoder position
+  const double incremental_velocity = (new_incremental_position - this->incremental_position_) / elapsed_time.toSec();
+  const double absolute_velocity = (new_absolute_position - this->absolute_position_) / elapsed_time.toSec();
+
+  if (std::abs(incremental_velocity - absolute_velocity) * elapsed_time.toSec() >
```
RutgerVanBeek: This if statement is quit unreadable. I think you should either add a comment on what it is doing or break it up so that the resulting if statement looks like `if position_difference > threshold`. Furthermore Some parts don't need to be calculated over and over again. The second part of the comparison is constant.

march_hardware/src/Joint.cpp
```diff
@@ -50,34 +55,62 @@ void Joint::actuateRad(double targetPositionRad)
...
 
-double Joint::getAngleRadAbsolute()
+void Joint::readEncoders(const ros::Duration& elapsed_time)
 {
-  if (!hasIMotionCube())
+  const double new_incremental_position = this->iMotionCube.getAngleRadIncremental();
+  const double new_absolute_position = this->iMotionCube.getAngleRadAbsolute();
+
+  // Get velocity from encoder position
+  const double incremental_velocity = (new_incremental_position - this->incremental_position_) / elapsed_time.toSec();
```
RutgerVanBeek: May be put the 'elapsed_time.toSec()' in a variable as you use it three time.

RutgerVanBeek (changes requested):
I am not sure whether this filtering technique is best for our problems. I will try to dig deeper into the movements of the incremental encoder today. Because this is only a problem during a joint in idle state we might consider adding the desired velocity into equation. Furthermore if one encoder is a lot more precise than the other (which we is our case) it can have noise without the other encoder changing a bit. The first if statement will fail, because the error is not bigger than 2 bits for the less precise encoder, so the filtering won't do anything, although this is a case we want to tackle.

march_hardware/test/encoder/TestIncrementalEncoder.cpp
Roelemans: There is :)

march_hardware/src/IMotionCube.cpp
Roelemans: Oops :sweat_smile:

Roelemans:
`Because this is only a problem during a joint in idle state we might consider adding the desired velocity into equation. `
Don't the encoders always have outliers?

`Furthermore if one encoder is a lot more precise than the other (which we is our case) it can have noise without the other encoder changing a bit. The first if statement will fail, because the error is not bigger than 2 bits for the less precise encoder, so the filtering won't do anything, although this is a case we want to tackle.`
I think that it is not possible to filter out noise of that magnitude, unless you use a smoothing which we would like to avoid. The purpose of this branch is more to filter out the outliers our sensor data occasionally has, than to filter out noise.

march_hardware/src/Joint.cpp
Roelemans: If things work as I and Bjorn hope, than recalibration should be very rare :) but we will see in the tests

march_hardware/src/Joint.cpp
Roelemans: elapsed time is already kind of a variable and toSec is similar to a getter, so it has my preference to leave it like this. Maybe one of the other software boys kan give a third opinion?

march_hardware/src/Joint.cpp
Roelemans: I added a comment :+1:

RutgerVanBeek:
I think the problem of the noise on the encoder data is not so big and I think this filter would not make it better. For what I can see in what we have on data on incremental encoders. There won't be outliers of this size. Therefore I think the filter should not be part of the PR. I think we can just use the incremental encoders for the position without additional filtering. 

Furthermore I think it is a good idea to add the absolute encoder value to the IMC msg. This way we can compare the value of both encoders and see if something can be improved in determining the position.

march_hardware/src/Joint.cpp
RutgerVanBeek: I looked at some data and indeed calibration should not happend often.

Roelemans:
Tested on exo and it does not work. Will write test functions first in the hopes of finding the problem.

Roelemans:
I wrote test functions and the code seems to work just fine. Perhaps there was a bug that was unwittingly fixed while merging one of the other resent PRs. Will test this on exo on Wednesday. If it appears the encoders do not have any outliers I will remove that part of the code (Joint.cpp line 92-108), or replace it by a simple warning that does nothing else.

Joris-JW (commented):
I thin the PR looks good. Wednesday u will have more information about merging this I expect. U also mentioned that our encoders could have noise in a static position. This seems to me like something we could measure? can we not also do this Wednesday?

march_hardware/include/march_hardware/encoder/Encoder.h
```diff
@@ -36,6 +36,13 @@ class Encoder
    */
   virtual double toRad(int32_t iu) const = 0;
 
+  /*
```
Olavhaasie: Necessary for docstrings
```suggestion
  /**
```

march_hardware/include/march_hardware/IMotionCube.h
```diff
@@ -40,9 +40,10 @@ class IMotionCube : public Slave
 
   void writeInitialSDOs(int cycle_time) override;
 
-  double getAngleRadAbsolute();
-  double getAngleRadIncremental();
-  double getAngleRadMostPrecise();
+  virtual double getAngleRadAbsolute();
+  virtual double getAngleRadIncremental();
+  double getAbsoluteRadPerBit();
+  double getIncrementalRadPerBit();
```
Olavhaasie: These functions can be `const`, since they only call a `const` function from incremental and absolute encoders. You should make functions `const` whenever possible: https://isocpp.org/wiki/faq/const-correctness

march_hardware/include/march_hardware/Joint.h
```diff
@@ -49,10 +49,12 @@ class Joint
...
   void actuateTorque(int16_t target_torque);
+  void readEncoders(const ros::Duration& elapsed_time);
 
-  double getAngleRadAbsolute();
-  double getAngleRadIncremental();
-  double getAngleRadMostPrecise();
+  double getPosition();
+  double getVelocity();
+  double getIncremental_position();
+  double getAbsolute_position();
```
Olavhaasie: You should make these methods `const`, since all they do is return a value without modifying the class. (Also use camelCase naming :stuck_out_tongue:)
```suggestion
  double getPosition() const;
  double getVelocity() const;
  double getIncrementalPosition() const;
  double getAbsolutePosition() const;
```

march_hardware/include/march_hardware/Joint.h
```diff
@@ -121,6 +123,11 @@ class Joint
   bool allow_actuation_ = false;
   float previous_motor_volt_ = 0.0;
 
+  double position_;
+  double incremental_position_;
+  double absolute_position_;
+  double velocity_;
```
Olavhaasie: You should initialize these members

RutgerVanBeek:
> I thin the PR looks good. Wednesday u will have more information about merging this I expect. U also mentioned that our encoders could have noise in a static position. This seems to me like something we could measure? can we not also do this Wednesday?

Both the incremental en absolute encoder data are in the imc state message. We thus have the raw encoder values, which we can use for analysis, also on noise indeed.

march_hardware/src/Joint.cpp
```diff
@@ -65,34 +70,94 @@ void Joint::actuateRad(double target_position)
...
+    const double new_incremental_position = this->imc_->getAngleRadIncremental();
+    const double new_absolute_position = this->imc_->getAngleRadAbsolute();
+
+    // Get velocity from encoder position
+    const double incremental_velocity = (new_incremental_position - this->incremental_position_) / elapsed_time.toSec();
+    const double absolute_velocity = (new_absolute_position - this->absolute_position_) / elapsed_time.toSec();
+
+    // If the difference in movement of the two encoders is larger than twice the sum of their precisions,
+    // one of them is probably an outlier.
+    if (std::abs(incremental_velocity - absolute_velocity) * elapsed_time.toSec() >
```
Olavhaasie: You should include the [`cmath`](https://en.cppreference.com/w/cpp/header/cmath) when using functions from the standard math library.

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -274,7 +257,7 @@ void MarchHardwareInterface::write(const ros::Time& /* time */, const ros::Durat
       }
       else if (joint.getActuationMode() == march::ActuationMode::torque)
       {
-        joint.actuateTorque(std::round(joint_effort_command_[i]));
+        joint.actuateTorque(0);
```
Olavhaasie, line 260: :thinking:

march_hardware/test/TestJoint.cpp
```diff
@@ -149,3 +149,166 @@ TEST_F(JointTest, GetTemperatureWithoutTemperatureGes)
...
+  march::Joint joint("actuate_true", 0, true, std::move(this->imc));
+  joint.receivedDataUpdate();
+  ASSERT_FALSE(joint.receivedDataUpdate());
+}
+
+TEST_F(JointTest, TestReadEncodersOnce)
+{
+  ros::Duration elapsed_time(0.2);
+  double velocity = 0.5;
+  double absolute_noise = -2 * this->imc->getAbsoluteRadPerBit();
```
Olavhaasie, line 191: You didn't mock the `getAbsoluteRadPerBit` function here, so what will this return?

march_hardware/src/Joint.cpp
```diff
@@ -65,34 +70,94 @@ void Joint::actuateRad(double target_position)
...
+
+    // Update position with the most accurate velocity
+    this->position_ += this->velocity_ * elapsed_time.toSec();
+    this->incremental_position_ = new_incremental_position;
+    this->absolute_position_ = new_absolute_position;
+  }
+  else
+  {
+    // Update positions with velocity from last time step
+    this->position_ += this->velocity_ * elapsed_time.toSec();
```
RutgerVanBeek, line 141: I don't now how big of a deal this is, but this could result in unnecessary floating point errors, because you are basically reversing one of the calculations above.

march_hardware/test/TestJoint.cpp
```diff
@@ -149,3 +149,166 @@ TEST_F(JointTest, GetTemperatureWithoutTemperatureGes)
...
+      .WillOnce(Return(initial_absolute_position))
+      .WillOnce(Return(new_absolute_position));
+
+  march::Joint joint("actuate_true", 0, true, std::move(this->imc));
+  joint.prepareActuation();
+
+  joint.readEncoders(elapsed_time);
+
+  ASSERT_NEAR(joint.getPosition(), initial_absolute_position + velocity * elapsed_time.toSec(), std::pow(10, -9));
+  ASSERT_NEAR(joint.getVelocity(), velocity, std::pow(10, -9));
```
Olavhaasie: There also exists a [`ASSERT_DOUBLE_EQ`](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#floating-point-comparison) which will check `double`s for equality within 4 Units in the Last Place (ULPs). Which will work for better for all value ranges and you don't have to give an error bound.

march_hardware/test/TestJoint.cpp
```diff
@@ -149,3 +149,166 @@ TEST_F(JointTest, GetTemperatureWithoutTemperatureGes)
...
+}
+
+TEST_F(JointTest, TestReadEncodersTwice)
+{
+  ros::Duration elapsed_time(0.2);
+  double first_velocity = 0.5;
+  double second_velocity = 0.8;
+
+  double absolute_noise = -this->imc->getAbsoluteRadPerBit();
```
Olavhaasie, line 222: Sam here, this method isn't mocked, so what do you expect?

march_hardware/test/encoder/TestIncrementalEncoder.cpp
```diff
@@ -22,5 +22,5 @@ TEST_F(TestIncrementalEncoder, CorrectToRad)
 {
   const int32_t iu = 1000;
   const double expected = iu * 2.0 * M_PI / (std::pow(2, this->resolution) * this->transmission);
-  ASSERT_EQ(expected, this->encoder.toRad(iu));
+  ASSERT_NEAR(expected, this->encoder.toRad(iu), std::pow(10, -9));
```
Olavhaasie: Same here, checkout `ASSERT_DOUBLE_EQ`

Olavhaasie (changes requested):

RutgerVanBeek (commented):
Nice testing!

march_hardware/include/march_hardware/IMotionCube.h
Roelemans: ~Same goes for getAngleRadAbsolute() and getAngleRadIncremental on line 43 and 44 no?~ I just found out you can't const those.

 Should I mark them as const again at their definition in the cpp file?

march_hardware/include/march_hardware/Joint.h
Roelemans: Oof to think that name slipped through :fearful:

march_hardware/include/march_hardware/Joint.h
Roelemans: They are "initialized" in Joint::prepareActuation, which should be called before anything else. But I guess such things should be initialized either in the header file (here) or in the constructor correct?

march_hardware_interface/src/march_hardware_interface.cpp
Roelemans: This is just for the test :sweat_smile: will be removed if the test proves succesful

march_hardware/test/TestJoint.cpp
Roelemans: Whatever the resolutions of the encoders are initialized to by default. joint.readEncoders uses this value to determine whether something is noise or an outlier, so for good testing I need to use that value. Whatever that value is is not important.

march_hardware/include/march_hardware/Joint.h
Olavhaasie: `Joint` has multiple constructors and since all values should be initialized the same, you can just do it here in the header file

march_hardware/test/TestJoint.cpp
Roelemans: Thanks that makes things better :)

march_hardware/src/Joint.cpp
Roelemans: Hmm good point. It is probably not that big a problem since you regularly re-calibrate anyway, but I changed it to use displacement instead of velocity.

PS tests were very handy to debug this change :smiley:

RutgerVanBeek (dismissed):

march_hardware/include/march_hardware/IMotionCube.h
Olavhaasie: You should mark functions `const` as much as possible, so if it's possible yes :+1:

Olavhaasie (dismissed):

Joris-JW (dismissed):
nice.

Roelemans:
Tested, doesn't work. Will reopen when the problem is fixed.

PR #243, opened 2020-03-23
Only update controllers if ethercat train has returned

Roelemans:
Closes PM-338

## Description

The EthercatMaster now signals the march_hardware_interface_node, which then  updates the controller. In addition, the read method of the march_hardware_interface checks for every joint if the data of that joint has been updated. If it has not, it keeps velocity constant and updates positions accordingly. 

## Changes

* Ethercat master uses std::condition_variable to send signal when PDO is received
* march_hardware_interface_node only updates controller when the signal is received
* Velocity is kept constant for every joint that did not receive an update.
* Use actual time since last update as input for velocity instead of expectedCycleTime

### DIFF
--- a/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
+++ b/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
@@ -33,6 +33,8 @@ public:
   EthercatMaster& operator=(EthercatMaster&&) = delete;
 
   bool isOperational() const;
+  bool getTrainReturned();
+  void setTrainReturned(bool train_returned);
 
   /**
    * Returns the cycle time in milliseconds.
@@ -83,6 +85,7 @@ private:
   const std::string ifname_;
   const int max_slave_index_;
   const int cycle_time_ms_;
+  bool train_returned_ = false;
 
   char io_map_[4096] = { 0 };
   int expected_working_counter_ = 0;
--- a/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
+++ b/march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
@@ -5,6 +5,8 @@
 #include <vector>
 #include <string>
 #include <thread>
+#include <mutex>
+#include <condition_variable>
 
 #include <march_hardware/Joint.h>
 
@@ -33,8 +35,7 @@ public:
   EthercatMaster& operator=(EthercatMaster&&) = delete;
 
   bool isOperational() const;
-  bool getTrainReturned();
-  void setTrainReturned(bool train_returned);
+  void waitForPdo();
 
   /**
    * Returns the cycle time in milliseconds.
@@ -85,7 +86,10 @@ private:
   const std::string ifname_;
   const int max_slave_index_;
   const int cycle_time_ms_;
-  bool train_returned_ = false;
+
+  std::mutex wait_on_pdo_condition_mutex_;
+  std::condition_variable wait_on_pdo_condition_var_;
+  bool pdo_received_ = false;
 
   char io_map_[4096] = { 0 };
   int expected_working_counter_ = 0;
--- a/march_hardware/include/march_hardware/Joint.h
+++ b/march_hardware/include/march_hardware/Joint.h
@@ -19,7 +19,7 @@ private:
   // Set this number via the hardware builder
   int netNumber;
   bool allowActuation;
-  float previous_motor_volt_;
+  float previous_motor_volt_ = 0.0;
   IMotionCube iMotionCube;
   TemperatureGES temperatureGES;
 
--- a/march_hardware/include/march_hardware/Joint.h
+++ b/march_hardware/include/march_hardware/Joint.h
@@ -19,6 +19,7 @@ private:
   // Set this number via the hardware builder
   int netNumber;
   bool allowActuation;
+  float previous_motor_volt_;
   IMotionCube iMotionCube;
   TemperatureGES temperatureGES;
 
@@ -56,6 +57,7 @@ public:
   bool hasIMotionCube();
   bool hasTemperatureGES();
   bool canActuate();
+  bool receivedDataUpdate();
 
   /** @brief Override comparison operator */
   friend bool operator==(const Joint& lhs, const Joint& rhs)
--- a/march_hardware/include/march_hardware/MarchRobot.h
+++ b/march_hardware/include/march_hardware/MarchRobot.h
@@ -50,8 +50,7 @@ public:
 
   bool isEthercatOperational();
 
-  bool getTrainReturned();
-  void setTrainReturned(bool train_returned);
+  void waitForPdo();
 
   int getEthercatCycleTime() const;
 
--- a/march_hardware/include/march_hardware/MarchRobot.h
+++ b/march_hardware/include/march_hardware/MarchRobot.h
@@ -50,6 +50,9 @@ public:
 
   bool isEthercatOperational();
 
+  bool getTrainReturned();
+  void setTrainReturned(bool train_returned);
+
   int getEthercatCycleTime() const;
 
   Joint getJoint(::std::string jointName);
--- a/march_hardware/src/EtherCAT/EthercatMaster.cpp
+++ b/march_hardware/src/EtherCAT/EthercatMaster.cpp
@@ -157,6 +157,7 @@ void EthercatMaster::ethercatLoop()
   {
     const auto begin_time = std::chrono::high_resolution_clock::now();
 
+    this->sendReceivePdo();
     monitorSlaveConnection();
 
     const auto end_time = std::chrono::high_resolution_clock::now();
--- a/march_hardware/src/EtherCAT/EthercatMaster.cpp
+++ b/march_hardware/src/EtherCAT/EthercatMaster.cpp
@@ -163,17 +163,18 @@ void EthercatMaster::ethercatLoop()
     const auto end_time = std::chrono::high_resolution_clock::now();
     const auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end_time - begin_time);
 
+    {
+      std::lock_guard<std::mutex> lock(this->wait_on_pdo_condition_mutex_);
+      this->pdo_received_ = true;
+    }
+    this->wait_on_pdo_condition_var_.notify_one();
+
     if (duration > cycle_time)
     {
       not_achieved_count++;
     }
     else
     {
-      {
-        std::lock_guard<std::mutex> lock(this->wait_on_pdo_condition_mutex_);
-        this->pdo_received_ = true;
-      }
-      this->wait_on_pdo_condition_var_.notify_one();
       std::this_thread::sleep_for(cycle_time - duration);
     }
     total_loops++;
--- a/march_hardware/src/EtherCAT/EthercatMaster.cpp
+++ b/march_hardware/src/EtherCAT/EthercatMaster.cpp
@@ -41,14 +41,12 @@ int EthercatMaster::getCycleTime() const
   return this->cycle_time_ms_;
 }
 
-bool EthercatMaster::getTrainReturned()
+void EthercatMaster::waitForPdo()
 {
-  return this->train_returned_;
-}
-
-void EthercatMaster::setTrainReturned(bool train_returned)
-{
-  this->train_returned_ = train_returned;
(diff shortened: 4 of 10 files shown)


RutgerVanBeek (changes requested):
Looks good. I only think the naming is misleading. Especially `setTrainReturned` because it is not intuitive whether the hardware interface decides that a train returned. I think a better variable name would be something like `lastTrainIsRead` and then reverse the logic.

march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
```diff
@@ -83,6 +85,7 @@ class EthercatMaster
   const std::string ifname_;
   const int max_slave_index_;
   const int cycle_time_ms_;
+  bool train_returned_ = false;
```
Olavhaasie: I think that `train` is not a very good descriptive name. This is not how it is represented in the SOEM documentation anyway. I would name it something like `received_pdo_` or `received_packets_`.

Furthermore, this member is written to/read from 2 different threads and should therefore be an `std::atomic<bool>`, same as `is_operational_`.

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -153,6 +163,8 @@ void EthercatMaster::ethercatLoop()
     this->sendReceivePdo();
     monitorSlaveConnection();
 
+    this->train_returned_ = true;
```
Olavhaasie: It could be possible that the timeout was reached and no PDOs were received. So, I guess you don't want to say that the 'train' returned

march_hardware/include/march_hardware/MarchRobot.h
```diff
@@ -50,6 +50,9 @@ class MarchRobot
 
   bool isEthercatOperational();
 
+  bool getTrainReturned();
```
Olavhaasie: This method can be `const`
```suggestion
  bool getTrainReturned() const;
```
(Same as all other getters you added)

march_hardware_interface/src/march_hardware_interface_node.cpp
```diff
@@ -45,16 +45,22 @@ int main(int argc, char** argv)
...
-      controller_manager.update(now, rate.expectedCycleTime());
-      march.write(now, rate.expectedCycleTime());
+      if (march.getTrainReturned())
+      {
+        march.setTrainReturned(false);
+        march.read(now, now - last_update_time);
+        last_update_time = now;
+        march.validate();
+        controller_manager.update(now, rate.expectedCycleTime());
+        march.write(now, rate.expectedCycleTime());
```
Olavhaasie: Shouldn't you change the cycle time here as well? The expected cycle time is always the theoretical cycle time. So, when the previous cycle took longer than it was supposed to, the cycle time will be incorrect.

march_hardware_interface/src/march_hardware_interface_node.cpp
```diff
@@ -45,16 +45,22 @@ int main(int argc, char** argv)
...
     const ros::Time now = ros::Time::now();
     try
     {
-      march.read(now, rate.expectedCycleTime());
-      march.validate();
-      controller_manager.update(now, rate.expectedCycleTime());
-      march.write(now, rate.expectedCycleTime());
+      if (march.getTrainReturned())
+      {
+        march.setTrainReturned(false);
```
Olavhaasie: Essentially what you want to do here is wait for new data to arrive. You can probably do this a lot more efficient using [`std::condition_variable`](https://en.cppreference.com/w/cpp/thread/condition_variable), so that you can be notified when a new train arrived. However, when implementing this I am wondering what the advantage then is of having the ethercat train is separate thread anyway, since now you are forcing it to be sequential, which you can do a lot easier using a single thread. What is the core reason that you want to implement this? What problem does it solve?

Olavhaasie (changes requested):

march_hardware_interface/src/march_hardware_interface_node.cpp
RutgerVanBeek: I think that ideally you want to update your controller as fast as there is new information. Although it could be argued how much it would matter.

Roelemans: You want to prevent the controller updating before there is a data update, because then it will think that the velocity is zero (and the velocity of the next timestep will be 2x the normal). I thought about putting them in the same thread, but I wanted to avoid the controller somehow gobbeling up computation and hence delaying the train. A signal using std::condition_variable seems like a good idea but little stupid me did not know how to do this. I will look at it tomorrow :)

march_hardware/include/march_hardware/EtherCAT/EthercatMaster.h
Roelemans: Changed the names. I believe the atomic<bool> is no longer necessary since I am using std::condition_variable now.

march_hardware/src/EtherCAT/EthercatMaster.cpp
Roelemans: Good point. I moved it to the else statement below

march_hardware_interface/src/march_hardware_interface_node.cpp
Roelemans: Oops

march_hardware_interface/src/march_hardware_interface_node.cpp
Roelemans: I implemented something but I am far from certain it will work :sweat_smile: Please have a look and bestow me your wisdom :)

march_hardware/include/march_hardware/MarchRobot.h
Roelemans: The getters no long exist :+1:

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -41,6 +41,14 @@ int EthercatMaster::getCycleTime() const
   return this->cycle_time_ms_;
 }
 
+void EthercatMaster::waitForPdo()
+{
+  std::unique_lock<std::mutex> lock(this->wait_on_pdo_condition_mutex_);
+  this->wait_on_pdo_condition_var_.wait(lock, [&] { return this->pdo_received_; });
+  this->pdo_received_ = false;
+  lock.unlock();
```
Olavhaasie: This unlock is not necessary since [`std::unique_lock`](http://www.cplusplus.com/reference/mutex/unique_lock/) automatically releases the lock when destroyed

RutgerVanBeek (commented):
I approve, although I cannot really judge on the c++ implementation on the lock. One question still, the hardware interface cycle time fully depends on ethercat now. Don't we may be want a minimum delay, because we don't exactly how the controller would react to really small elapsed times? One bit encoder movements can cause huge spikes in velocity, making the controller unstable.

march_hardware_interface/src/march_hardware_interface_node.cpp
```diff
@@ -41,21 +41,23 @@ int main(int argc, char** argv)
     return 1;
   }
 
-  ros::Rate rate(
-      ros::Duration(march.getEthercatCycleTime() * ros::param::param<int>("~control_loop_multiplier", 1) / 1000.0));
```
Olavhaasie, line 45: You should also remove the parameter from the launch file

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -162,6 +170,9 @@ void EthercatMaster::ethercatLoop()
     }
     else
     {
+      std::lock_guard<std::mutex> lock(this->wait_on_pdo_condition_mutex_);
+      this->pdo_received_ = true;
+      this->wait_on_pdo_condition_var_.notify_one();
```
Olavhaasie: Why only notify when a cycle was achieved? It could be that the cycle took longer but still received its PDOs, due to OS scheduling. I would let `sendReceivePdo` return wether the pdos were received and only then notify the cv.

And small note, you have to unlock the mutex before [`notify_one`](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one) or the cv will wake up and immediately go back to sleep.
```suggestion
      std::lock_guard<std::mutex> lock(this->wait_on_pdo_condition_mutex_);
      this->pdo_received_ = true;
      lock.unlock();
      this->wait_on_pdo_condition_var_.notify_one();
```

Olavhaasie (changes requested):

march_hardware/src/EtherCAT/EthercatMaster.cpp
Roelemans: Oh you are right, I misinterpreted the example.

march_hardware/src/EtherCAT/EthercatMaster.cpp
Roelemans: Hmm, is this unnecessary in the example because it automatically gets destroyed when it exists {}?
image

Edit: apparently the lock_guard does not have an unlock. It must be destructed to unlock the mutex. Hence, I put it inside a block like in this example instead.

Olavhaasie: Yes

Roelemans: I now only send a signal if the wkc is 29. Probably not ideal since you stop updating the controller when you loose a slave, but we saw in the logs that sometimes only a part of the slaves send data, probably due to wkc too low. Ideally, of course, you would like to then update only those slaves from whom you received data. That is a bit complicated in the current format, however :/.

Olavhaasie: Hmm this is getting a bit complicated, since you want to keep updating all slaves. Say for example that every cycle another slave loses connection, first 1 then 2,3,... Then no slave gets updated :frowning_face:

Roelemans: Perhaps we have to go for/combine this with another approach after all. I can create a  check within the joint class whether the parameters have been updated (for example by assuming the the motor voltage never completely remains the same). I can then call this check for every joint separately in the loop of MarchHardwareInterface::read.

Olavhaasie: Okay, let's close this PR for now

Olavhaasie:
Closed to work on another approach. See the [thread](https://github.com/project-march/hardware-interface/pull/243#discussion_r397047384)

Roelemans:
I added a method to the joint that checks whether the motor voltage has changed since the previous request, assuming the PDO has updated its data if and only if the motor voltage has changed. The march_hardware_interface::read then keeps velocity constant and updates positions accordingly if a joint's data has not been updated.

march_hardware/include/march_hardware/Joint.h
```diff
@@ -19,6 +19,7 @@ class Joint
   // Set this number via the hardware builder
   int netNumber;
   bool allowActuation;
+  float previous_motor_volt_;
```
Olavhaasie: You should initialize this variable to a value (probably `0.0`)

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -162,6 +169,11 @@ void EthercatMaster::ethercatLoop()
     }
     else
     {
+      {
+        std::lock_guard<std::mutex> lock(this->wait_on_pdo_condition_mutex_);
+        this->pdo_received_ = true;
+      }
+      this->wait_on_pdo_condition_var_.notify_one();
```
Olavhaasie: Again, same question as before, why do you only notify when the cycle was achieved? It could be that the cycle could not be achieved due to OS scheduling or longer transport times, but the new data arrived anyway.

Olavhaasie (changes requested):

march_hardware/src/EtherCAT/EthercatMaster.cpp
Roelemans: Fixed while you were reviewing :sweat_smile:

march_hardware/include/march_hardware/Joint.h
Roelemans: Oops wanted to do that at one point but forgot

Olavhaasie (approved):

RutgerVanBeek (approved):

PR #244, opened 2020-03-24
Fix: dependency injection

Olavhaasie:
Closes PM-341

## Description
The mocks were used in the wrong way. They were being passed by value to their base classes which resulted in the mocked functions to be scraped away. So, for example, a function takes `Encoder` as type but you pass a `MockEncoder` which inherits from `Encoder`. The method only has space on its stack for an `Encoder` and since `MockEncoder` implements some overridden methods it does not fit in the memory space of `Encoder`. Therefore, the mocks were not used. Luckily this can be fixed by passing pointers. I implemented this using smart pointers, e.g. `std::unique_ptr`. This type of pointer destroys itself and ensures there is only one instance. This way you don't have to think about `delete`'ing pointers and worry about ownership.

Since pointers can also be `nullptr`, you have to check everywhere if a pointer is valid before using it. In `IMotionCube` class I solved this using a check on the encoders in the constructor, since these are required. In Joint I changed the check to allow for actuation to also check for an existing motor controller.

Finally, all further additions are tests using the mocks correctly (to increase the coverage).

## Changes
* Make `AbsoluteEncoder` and `IncrementalEncoder` injectable into `IMotionCube`
* Make `IMotionCube` and `TemperatureGes` injectable into `Joint`
* Write additional tests for `IMotionCube` and `Joint`
* Make `getJoint` in `MarchRobot` return a reference instead of a value to avoid copying.
* Refactor `Joint` a bit to make it less mutable and add constness (fixes PM-324)
* Make destructors `virtual` to enable for correctly destructing of inherited classes, e.g. mocks

### DIFF
--- a/march_hardware/include/march_hardware/IMotionCube.h
+++ b/march_hardware/include/march_hardware/IMotionCube.h
@@ -97,7 +97,7 @@ private:
 
   // Use of smart pointers are necessary here to make dependency injection
   // possible and thus allow for mocking the encoders. A unique pointer is
-  // is chosen since the IMotionCube should be the owner and the encoders
+  // chosen since the IMotionCube should be the owner and the encoders
   // do not need to be passed around.
   std::unique_ptr<AbsoluteEncoder> absolute_encoder_;
   std::unique_ptr<IncrementalEncoder> incremental_encoder_;
--- a/march_hardware/include/march_hardware/IMotionCube.h
+++ b/march_hardware/include/march_hardware/IMotionCube.h
@@ -58,11 +58,11 @@ public:
 
   void setControlWord(uint16_t control_word);
 
-  void actuateRad(double target_rad);
-  void actuateTorque(int16_t target_torque);
+  virtual void actuateRad(double target_rad);
+  virtual void actuateTorque(int16_t target_torque);
 
   void goToTargetState(IMotionCubeTargetState target_state);
-  void goToOperationEnabled();
+  virtual void goToOperationEnabled();
 
   /** @brief Override comparison operator */
   friend bool operator==(const IMotionCube& lhs, const IMotionCube& rhs)
--- a/march_hardware/include/march_hardware/IMotionCube.h
+++ b/march_hardware/include/march_hardware/IMotionCube.h
@@ -32,7 +32,11 @@ public:
   IMotionCube(int slave_index, std::unique_ptr<AbsoluteEncoder> absolute_encoder,
               std::unique_ptr<IncrementalEncoder> incremental_encoder, ActuationMode actuation_mode);
 
-  ~IMotionCube() = default;
+  ~IMotionCube() noexcept override = default;
+
+  /* Delete copy constructor/assignment since the unique_ptrs cannot be copied */
+  IMotionCube(const IMotionCube&) = delete;
+  IMotionCube& operator=(const IMotionCube&) = delete;
 
   void writeInitialSDOs(int cycle_time) override;
 
--- a/march_hardware/include/march_hardware/IMotionCube.h
+++ b/march_hardware/include/march_hardware/IMotionCube.h
@@ -11,6 +11,7 @@
 #include "march_hardware/PDOmap.h"
 #include "march_hardware/Slave.h"
 
+#include <memory>
 #include <unordered_map>
 #include <string>
 
@@ -19,8 +20,17 @@ namespace march
 class IMotionCube : public Slave
 {
 public:
-  IMotionCube(int slave_index, AbsoluteEncoder absolute_encoder, IncrementalEncoder incremental_encoder,
-              ActuationMode actuation_mode);
+  /**
+   * Constructs an IMotionCube with an incremental and absolute encoder.
+   *
+   * @param slave_index index of the IMotionCube slave
+   * @param absolute_encoder pointer to absolute encoder, required so cannot be nullptr
+   * @param incremental_encoder pointer to incremental encoder, required so cannot be nullptr
+   * @param actuation_mode actuation mode in which the IMotionCube must operate
+   * @throws std::invalid_argument When an absolute or incremental encoder is nullptr.
+   */
+  IMotionCube(int slave_index, std::unique_ptr<AbsoluteEncoder> absolute_encoder,
+              std::unique_ptr<IncrementalEncoder> incremental_encoder, ActuationMode actuation_mode);
 
   ~IMotionCube() = default;
 
@@ -60,8 +70,8 @@ public:
   friend std::ostream& operator<<(std::ostream& os, const IMotionCube& imc)
   {
     return os << "slaveIndex: " << imc.slaveIndex << ", "
-              << "incrementalEncoder: " << imc.incremental_encoder_ << ", "
-              << "absoluteEncoder: " << imc.absolute_encoder_;
+              << "incrementalEncoder: " << imc.incremental_encoder_.get() << ", "
+              << "absoluteEncoder: " << imc.absolute_encoder_.get();
   }
 
   constexpr static double MAX_TARGET_DIFFERENCE = 0.393;
@@ -81,8 +91,12 @@ private:
   void mapMosiPDOs();
   void writeInitialSettings(uint8_t cycle_time);
 
-  AbsoluteEncoder absolute_encoder_;
-  IncrementalEncoder incremental_encoder_;
+  // Use of smart pointers are necessary here to make dependency injection
+  // possible and thus allow for mocking the encoders. A unique pointer is
+  // is chosen since the IMotionCube should be the owner and the encoders
+  // do not need to be passed around.
+  std::unique_ptr<AbsoluteEncoder> absolute_encoder_;
+  std::unique_ptr<IncrementalEncoder> incremental_encoder_;
   ActuationMode actuation_mode_;
   bool is_incremental_more_precise_;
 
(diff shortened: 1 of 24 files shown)


march_hardware/src/IMotionCube.cpp
```diff
@@ -118,16 +125,17 @@ void IMotionCube::actuateRad(double target_rad)
...
 void IMotionCube::actuateIU(int32_t target_iu)
 {
-  if (!this->absolute_encoder_.isValidTargetIU(this->getAngleIUAbsolute(), target_iu))
+  if (!this->absolute_encoder_->isValidTargetIU(this->getAngleIUAbsolute(), target_iu))
   {
-    throw error::HardwareException(
-        error::ErrorType::INVALID_ACTUATE_POSITION, "Position %i is invalid for slave %d. (%d, %d)", target_iu,
-        this->slaveIndex, this->absolute_encoder_.getLowerSoftLimitIU(), this->absolute_encoder_.getUpperSoftLimitIU());
+    throw error::HardwareException(error::ErrorType::INVALID_ACTUATE_POSITION,
+                                   "Position %i is invalid for slave %d. (%d, %d)", target_iu, this->slaveIndex,
```
Roelemans: Why is the first %i whereas the other three are %d?

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -48,36 +49,37 @@ std::unique_ptr<march::MarchRobot> HardwareBuilder::createMarchRobot()
...
-  std::string robot_name = this->robot_config_.begin()->first.as<std::string>();
+  const auto robot_name = this->robot_config_.begin()->first.as<std::string>();
   ROS_DEBUG_STREAM("Starting creation of robot " << robot_name);
 
   // Remove top level robot name key
   YAML::Node config = this->robot_config_[robot_name];
-  std::string if_name = config["ifName"].as<std::string>();
-  int cycle_time = config["ecatCycleTime"].as<int>();
+  const auto if_name = config["ifName"].as<std::string>();
+  const auto cycle_time = config["ecatCycleTime"].as<int>();
```
Roelemans, line 58: Why use all these auto's? Wouldn't the code be clearer if you just use string and int here?

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -87,63 +89,71 @@ march::Joint HardwareBuilder::createJoint(const YAML::Node& joint_config, const
...
+    ROS_WARN("Joint %s does not have a configuration for an IMotionCube", joint_name.c_str());
+  }
+
+  auto ges = HardwareBuilder::createTemperatureGES(joint_config["temperatureges"]);
+  if (!ges)
   {
     ROS_WARN("Joint %s does not have a configuration for a TemperatureGes", joint_name.c_str());
   }
-  return joint;
+  return { joint_name, net_number, allow_actuation, std::move(imc), std::move(ges) };
```
Roelemans, line 121: Why do you use std::make_unique<march::...> for the other create methods but not for this one?

march_hardware/include/march_hardware/IMotionCube.h
```diff
@@ -81,8 +95,12 @@ class IMotionCube : public Slave
   void mapMosiPDOs();
   void writeInitialSettings(uint8_t cycle_time);
 
-  AbsoluteEncoder absolute_encoder_;
-  IncrementalEncoder incremental_encoder_;
+  // Use of smart pointers are necessary here to make dependency injection
+  // possible and thus allow for mocking the encoders. A unique pointer is
+  // is chosen since the IMotionCube should be the owner and the encoders
```
RutgerVanBeek:
```suggestion
  // possible and thus allow for mocking the encoders. A unique pointer is
  // chosen since the IMotionCube should be the owner and the encoders
```

RutgerVanBeek (changes requested):
Looks very nice, was really necessary to improve the ability for testing in the hardware interface. One really small comment.

march_hardware_builder/test/joint_builder_test.cpp
```diff
@@ -40,20 +40,18 @@ TEST_F(JointTest, ValidJointHip)
   this->joint->safety->soft_lower_limit = 0.1;
   this->joint->safety->soft_upper_limit = 1.9;
 
-  march::Joint created = HardwareBuilder::createJoint(config, "test_joint_hip", this->joint);
+  const std::string name = "test_joint_hip";
+  march::Joint created = HardwareBuilder::createJoint(config, name, this->joint);
+
+  auto absolute_encoder = std::make_unique<march::AbsoluteEncoder>(
+      16, 22134, 43436, this->joint->limits->lower, this->joint->limits->upper, this->joint->safety->soft_lower_limit,
```
Roelemans, line 47: Store these values in a parameter since you use them often?

march_hardware_builder/test/joint_builder_test.cpp
```diff
@@ -40,20 +40,18 @@ TEST_F(JointTest, ValidJointHip)
...
+      this->joint->safety->soft_upper_limit);
+  auto incremental_encoder = std::make_unique<march::IncrementalEncoder>(12, 50.0);
```
Roelemans, line 49: Same here

Roelemans (changes requested):
A few questions on c++ style and one on magic numbers in tests

march_hardware/src/IMotionCube.cpp
Olavhaasie: Good point, both %i and %d work for signed integers. I will change the %i to %d, since we use that everywhere else

march_hardware_builder/src/hardware_builder.cpp
Olavhaasie: I changed it to `auto` since the type is already contained in the `as<>()` call. And having the type double would be more work to refactor later.

march_hardware_builder/src/hardware_builder.cpp
Olavhaasie: Since the `MarchRobot` does not accept pointers to `Joint`s. When I think about it this also makes the joints not injectable with mocks. This is probably because the `MarchRobot` does not have any tests. I will add it to `MarchRobot`

march_hardware_builder/test/joint_builder_test.cpp
Olavhaasie: These values are from the test yaml and are only valid here, so I don't think it's necessary to turn them into parameters

march_hardware_builder/src/hardware_builder.cpp
Roelemans: Ah I see. This might not be necessary since Joint currently does not directly use any hardware, which should probably be kept this way in the future. It is enough to inject Joint with mock IMCs and mock Encoders

march_hardware_builder/test/joint_builder_test.cpp
Roelemans: Why not? That way, you will only have to change the value once when it changes in the yaml, and it is also clearer what these numbers stand for, since you would have to go look for the constructor to find out the first is bits, the second and third limits in IU.

march_hardware_builder/src/hardware_builder.cpp
Olavhaasie: Okay I agree

march_hardware_builder/test/joint_builder_test.cpp
Olavhaasie: Okay, so something like this?
```cpp
TEST_F(JointTest, ValidNotActuated)
{
  YAML::Node config = this->loadTestYaml("/joint_correct_not_actuated.yaml");
  this->joint->limits->lower = 0.0;
  this->joint->limits->upper = 2.0;
  this->joint->safety->soft_lower_limit = 0.1;
  this->joint->safety->soft_upper_limit = 1.9;
  const std::string name = "test_joint_hip";
  const size_t abs_encoder_resolution = 16;
  const int32_t abs_encoder_lower_limit = 22134;
  const int32_t abs_encoder_upper_limit = 43436;
  const size_t inc_encoder_resolution = 12;
  const double inc_encoder_transmission = 50.0;
  const int slave_index = 1;
  const int byte_offset = 2;
  const int net_number = -1;
  const bool allow_actuation = false;

  march::Joint created = HardwareBuilder::createJoint(config, name, this->joint);

  auto absolute_encoder = std::make_unique<march::AbsoluteEncoder>(
      abs_encoder_resolution, abs_encoder_lower_limit, abs_encoder_upper_limit, this->joint->limits->lower,
      this->joint->limits->upper, this->joint->safety->soft_lower_limit, this->joint->safety->soft_upper_limit);
  auto incremental_encoder =
      std::make_unique<march::IncrementalEncoder>(inc_encoder_resolution, inc_encoder_transmission);
  auto imc = std::make_unique<march::IMotionCube>(slave_index, std::move(absolute_encoder),
                                                  std::move(incremental_encoder), march::ActuationMode::unknown);
  auto ges = std::make_unique<march::TemperatureGES>(slave_index, byte_offset);
  march::Joint expected(name, net_number, allow_actuation, std::move(imc), std::move(ges));

  ASSERT_EQ(expected, created);
}
```

Roelemans: No, I meant put them in the setup since you use these values in most of the tests. The yaml is also loaded there so it makes sense to keep these together.

Roelemans (dismissed):

RutgerVanBeek (dismissed):

RutgerVanBeek (approved):

Roelemans (approved):

PR #246, opened 2020-03-25
Feature: increase ethercat thread priority

Olavhaasie:
Closes PM-311

## Description
In this PR I have researched the ethercat thread cycle time, to see if the ethercat thread should be given a higher priority or different scheduling priority.

In order to measure this I created a seperate branch,  [`time-ethercat-cycle`](https://github.com/project-march/hardware-interface/tree/time-ethercat-cycle), which measured ethercat thread timing. It measured the time it took to send and receive PDOs and it measured how long the ethercat cycle took in its entirety (including sleeping), which should always be 4ms. From this data I created histograms:

cycles
total

The first plot shows the sending and receiving of pdos and the second total duration. The y-axis are logarithmic. As you can see in the total duration there are a lot of cycles that take longer than 5ms.

Then I gave the ethercat thread more priority and gave it a different scheduling policy: `SCHED_FIFO`. This scheduling policy preempts other threads with lower priority as soon as it becomes runnable, i.e. wakes up from sleep. The priority has been set to 40 to give it more priority than other ROS nodes and not too high to avoid blocking OS stuff. Then I did the measurement again and got these results:

cycles_with_prio
total_with_prio

The ethercat sending and receiving PDOs now stays well under 2ms and the total duration is concentrated at 4ms, with many fewer outliers. As this improved the cycle times significantly, I propose to make this the default.

## Changes
* Add `setThreadPriority` method to set the ethercat thread priority

### DIFF
--- a/march_hardware/src/EtherCAT/EthercatMaster.cpp
+++ b/march_hardware/src/EtherCAT/EthercatMaster.cpp
@@ -7,6 +7,7 @@
 #include <string>
 #include <vector>
 
+#include <pthread.h>
 #include <ros/ros.h>
 
 #include <soem/ethercattype.h>
@@ -119,9 +120,7 @@ void EthercatMaster::ethercatSlaveInitiation(std::vector<Joint>& joints)
     ROS_INFO("Operational state reached for all slaves");
     this->is_operational_ = true;
     this->ethercat_thread_ = std::thread(&EthercatMaster::ethercatLoop, this);
-    struct sched_param param;
-    param.sched_priority = 40;
-    pthread_setschedparam(this->ethercat_thread_.native_handle(), SCHED_FIFO, &param);
+    this->setThreadPriority(EthercatMaster::THREAD_PRIORITY);
   }
   else
   {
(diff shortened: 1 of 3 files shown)


march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -217,4 +219,19 @@ void EthercatMaster::stop()
...
+  // SCHED_FIFO scheduling preempts other threads with lower priority as soon as it becomes runnable.
+  // See http://man7.org/linux/man-pages/man7/sched.7.html for more info.
+  const int error = pthread_setschedparam(this->ethercat_thread_.native_handle(), SCHED_FIFO, &param);
+  if (error != 0)
+  {
+    ROS_ERROR("Failed to set the ethercat thread priority to %d. (error code: %d)", priority, error);
+  }
+  else
+  {
+    ROS_DEBUG("Set ethercat thread priority to %d", param.sched_priority);
```
Roelemans, line 234: Do we ever look at debugs? They don't get logged either right?

RutgerVanBeek (approved):
What are the values for the other ROS nodes, OS systems and other applications, like Clion or Firefox?

march_hardware/src/EtherCAT/EthercatMaster.cpp
Olavhaasie: Not by default no, but if you want to you can :point_right:

Olavhaasie:
> What are the values for the other ROS nodes, OS systems and other applications, like Clion or Firefox?

The default is a static priority of 0 for user processes and uses `SCHED_OTHER` scheduling policy (http://man7.org/linux/man-pages/man7/sched.7.html). This is a time-sharing scheduler, which gives every thread the about the same amount of time. However, they can be given more priority with a _nice_ value.

The ethercat thread has a `SCHED_FIFO` scheduler, which means it will always preempt `SCHED_OTHER`. This is not much of a problem for the ethercat thread since it has a  very short execution time before it goes back to sleep (around 100us), but it needs to run on a tight schedule.

Roelemans (approved):

PR #247, opened 2020-03-26
Fix: wrong motor voltage reading

Ciliaclaij:
Part of
PM-300-add-motor-voltage

## Description

This pull request contains the first step to add the motor voltage. A mistake was made before: there was a reading of motor voltage, but it actually read the IMC voltage. This was changed. Furthermore object 2066h was added to the PDO mapping.

## Changes

* Renamed motor_voltage to IMC_voltage
* Added 2066h to PDO mapping

### DIFF
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -233,7 +233,7 @@ float IMotionCube::getMotorCurrent()
          static_cast<float>(motor_current_iu);  // Conversion to Amp, see Technosoft CoE programming manual
 }
 
-float IMotionCube::getMotorVoltage()
+float IMotionCube::getIMCVoltage()
 {
   const float V_DC_MAX_MEASURABLE = 102.3;    // maximum measurable DC voltage found in EMS Setup/Drive info button
   const float IU_CONVERSION_CONST = 65520.0;  // Conversion parameter, see Technosoft CoE programming manual
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -243,8 +243,10 @@ float IMotionCube::getMotorCurrent()
 
 float IMotionCube::getIMCVoltage()
 {
-  const float V_DC_MAX_MEASURABLE = 102.3;    // maximum measurable DC voltage found in EMS Setup/Drive info button
-  const float IU_CONVERSION_CONST = 65520.0;  // Conversion parameter, see Technosoft CoE programming manual (2015 page 89)
+  // maximum measurable DC voltage found in EMS Setup/Drive info button
+  const float V_DC_MAX_MEASURABLE = 102.3;
+  // Conversion parameter, see Technosoft CoE programming manual (2015 page 89)
+  const float IU_CONVERSION_CONST = 65520.0;
 
   uint16_t imc_voltage_iu =
       get_input_bit16(this->slaveIndex, this->miso_byte_offsets_.at(IMCObjectName::DCLinkVoltage)).ui;
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -238,10 +238,10 @@ float IMotionCube::getIMCVoltage()
   const float V_DC_MAX_MEASURABLE = 102.3;    // maximum measurable DC voltage found in EMS Setup/Drive info button
   const float IU_CONVERSION_CONST = 65520.0;  // Conversion parameter, see Technosoft CoE programming manual
 
-  uint16_t motor_voltage_iu =
+  uint16_t imc_voltage_iu =
       get_input_bit16(this->slaveIndex, this->miso_byte_offsets_.at(IMCObjectName::DCLinkVoltage)).ui;
   return (V_DC_MAX_MEASURABLE / IU_CONVERSION_CONST) *
-         static_cast<float>(motor_voltage_iu);  // Conversion to Volt, see Technosoft CoE programming manual
+         static_cast<float>(imc_voltage_iu);  // Conversion to Volt, see Technosoft CoE programming manual
 }
 
 void IMotionCube::setControlWord(uint16_t control_word)
--- a/march_hardware/src/IMotionCube.cpp
+++ b/march_hardware/src/IMotionCube.cpp
@@ -324,6 +324,13 @@ void IMotionCube::goToOperationEnabled()
   this->goToTargetState(IMotionCubeTargetState::OPERATION_ENABLED);
 }
 
+void IMotionCube::reset()
+{
+  this->setControlWord(0);
+  ROS_DEBUG("Slave: %d, Try to reset IMC", this->slaveIndex);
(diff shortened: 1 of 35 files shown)


march_hardware/src/PDOmap.cpp
```diff
@@ -21,6 +21,7 @@ std::unordered_map<IMCObjectName, IMCObject> PDOmap::all_objects = {
   { IMCObjectName::TargetPosition, IMCObject(0x607A, 32) },
   { IMCObjectName::TargetTorque, IMCObject(0x6071, 16) },
   { IMCObjectName::QuickStopDeceleration, IMCObject(0x6085, 32) },
+  { IMCObjectName::MotorVoltage, IMCObject(0x2066, 32) },
```
RPKrijnen: I believe 2066 is an SDO object, which doesn't need to be mapped.

march_hardware/src/Joint.cpp
```diff
@@ -140,7 +140,7 @@ IMotionCubeState Joint::getIMotionCubeState()
   states.motionErrorDescription = error::parseMotionError(this->iMotionCube.getMotionError());
 
   states.motorCurrent = this->iMotionCube.getMotorCurrent();
-  states.motorVoltage = this->iMotionCube.getMotorVoltage();
+  states.IMCVoltage = this->iMotionCube.getIMCVoltage();
```
RPKrijnen: As stated in my other comment, I believe that you are using an SDO object which does not support PDO mapping and therefore cannot make use of the same get-function. In PM-381 I will provide read functions for sdo objects, which should allow you to make use of 2065h and 2066h to read the motor voltage appropriately.

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -411,7 +411,7 @@ void MarchHardwareInterface::updateIMotionCubeState()
     imc_state_pub_->msg_.detailed_error_description[i] = imc_state.detailedErrorDescription;
     imc_state_pub_->msg_.motion_error_description[i] = imc_state.motionErrorDescription;
     imc_state_pub_->msg_.motor_current[i] = imc_state.motorCurrent;
-    imc_state_pub_->msg_.motor_voltage[i] = imc_state.motorVoltage;
+    imc_state_pub_->msg_.IMC_voltage[i] = imc_state.IMCVoltage;
```
RutgerVanBeek: In the march repository the message is defined to have a motor_voltage component, so you would have to change the message structure if you want this.

march_hardware/src/Joint.cpp
Roelemans: This is a renaming of the old method to read 0x2055, the IMC's input voltage. Cilia will probably add a getter for the motor voltage later :)

march_hardware/src/PDOmap.cpp
Roelemans: It would be neater if we could add this to the PDO like here, so maybe find out if this is possible through testing?

Joris-JW: Actually, this register cannot be mapped at all according to the IMC CoE manual. (see page 236)

march_hardware/src/IMotionCube.cpp
```diff
@@ -233,7 +233,7 @@ float IMotionCube::getMotorCurrent()
          static_cast<float>(motor_current_iu);  // Conversion to Amp, see Technosoft CoE programming manual
 }
 
-float IMotionCube::getMotorVoltage()
+float IMotionCube::getIMCVoltage()
 {
   const float V_DC_MAX_MEASURABLE = 102.3;    // maximum measurable DC voltage found in EMS Setup/Drive info button
   const float IU_CONVERSION_CONST = 65520.0;  // Conversion parameter, see Technosoft CoE programming manual
```
Joris-JW: page references would be nice

Roelemans (changes requested):
I think these changes are mostly fine :), but we cannot test it yet: you are currently still using getMotorVoltage in Joint::getIMotionCubeState (Joint.cpp line 229) which no longer exists. To make it testable I would implement a new getMotorVolgate() method that reads out Joint::getIMotionCubeState. Then set 2064 in IMotionCube::writeInitialSettings. Then it should be testable on wednesday during the software test :). Then we can find out whether, as @Deakensbubbles suggests, this doesn't work because 2066 is an SDO object.

march_hardware/src/Joint.cpp
Joris-JW: So the function getIMCVoltage does not yet exist?

Joris-JW (commented):
I think rogier has a valid point in first waiting for his SDO-read function to extract certain data before this PR can be used. Furthermore, since the current 2066 register is not possible using a PDO is register 2055 not exactly what u are looking for? _The object 2055 contains the actual value of the DC-link voltage. The object is expressed in internal voltage units._ ? let me know if this of any good!

march_hardware_interface/src/march_hardware_interface.cpp
Ciliaclaij: I actually already did this, because I saw my build failing. However I was not aware that I needed to make a seperate PR for the march repo, so I will do this asap!

march_hardware/src/Joint.cpp
Roelemans: No it does. getMotorVoltage no longer exists because she renamed it to getIMCVoltage

Ciliaclaij: Maybe it's good to clear some things up. I already made this branch to do add the Motor Voltage, but I realized that there were more steps. That's why I decided to make smaller pull requests, to make it easier and faster for you to comment my work (and for me to get trough the feedback). But I only have one Jira issue and one branch, so I did it this way and plan to continue to work on this Branch. I am not sure if this is they right way to do it, so if you guys think there is a better way to do this I like to hear it

march_hardware/src/PDOmap.cpp
Ciliaclaij: I realized this, so I think it's better to remove the register from the PDO mapping and use the read function Rogier will make

RPKrijnen (dismissed):
This is a nice PR for reading out the DC link voltage of the IMC. I have no further comments about the general functionality.

march_hardware/src/IMotionCube.cpp
Olavhaasie: Why didn't you comment on this/add this?

Ciliaclaij: I thought I was not the right person to do this, but I can look into it... I saw it as a 'note to self'

Roelemans (dismissed):
Good PR, nice and small :cat:

Roelemans:
Nevermind your clang build (checks on style) fails oops. You need to run
`find . -name '*.h' -or -name '*.cpp' | xargs clang-format -i -style=file`
Inside the hardware-interface folder in order to fix you format so that the clang build succeeds.

Joris-JW (dismissed):
If u fix the build merging would be fine.

Roelemans (approved):

RutgerVanBeek (approved):

PR #248, opened 2020-03-27
Feature/pm 381 add sdo read functions

RPKrijnen:
SDO read functions are now necessary for IMC functionality we did not use before.

## Description

The SDO read functions have been added in the same fashion as the SDO write functions were already. This PR also applies the new naming to old uses of the SDO write function.
The new functions have been tested on the testjoint and provided expected results i.e. changing the mode of operation SDO provided different, but expected, read values.
## Changes

- Added sdo_8/16/32bit_read functions to the EthercatSDO files
- Renamed sdo_8/16/32bit() to sdo_8/16/32bit_write()

### DIFF
--- a/march_hardware/src/EtherCAT/EthercatSDO.cpp
+++ b/march_hardware/src/EtherCAT/EthercatSDO.cpp
@@ -26,7 +26,7 @@ int sdo_bit8_write(uint16_t slave, uint16_t index, uint8_t sub, uint8_t value)
   return received_working_counter;
 }
 
-int sdo_bit8_read(uint16_t slave, uint16_t index, uint8_t sub, int *val_size, uint8_t *value)
+int sdo_bit8_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, uint8_t* value)
 {
   ROS_DEBUG("sdo_bit8: slaveIndex %i, reg 0x%X, subindex %i, value 0x%X", slave, index, sub, value);
   int received_working_counter = ec_SDOread(slave, index, sub, FALSE, val_size, value, EC_TIMEOUTRXM);
@@ -50,7 +50,7 @@ int sdo_bit16_write(uint16_t slave, uint16_t index, uint8_t sub, uint16_t value)
   return received_working_counter;
 }
 
-int sdo_bit16_read(uint16_t slave, uint16_t index, uint8_t sub, int *val_size, uint16_t *value)
+int sdo_bit16_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, uint16_t* value)
 {
   ROS_DEBUG("sdo_bit16: slaveIndex %i, reg 0x%X, subindex %i, value 0x%X", slave, index, sub, value);
   int received_working_counter = ec_SDOread(slave, index, sub, FALSE, val_size, value, EC_TIMEOUTRXM);
@@ -74,7 +74,7 @@ int sdo_bit32_write(uint16_t slave, uint16_t index, uint8_t sub, uint32_t value)
   return received_working_counter;
 }
 
-int sdo_bit32_read(uint16_t slave, uint16_t index, uint8_t sub, int *val_size, uint32_t *value)
+int sdo_bit32_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, uint32_t* value)
 {
   ROS_DEBUG("sdo_bit32: slaveIndex %i, reg 0x%X, subindex %i, value 0x%X", slave, index, sub, value);
   int received_working_counter = ec_SDOread(slave, index, sub, FALSE, val_size, value, EC_TIMEOUTRXM);
--- a/march_hardware/src/EtherCAT/EthercatSDO.cpp
+++ b/march_hardware/src/EtherCAT/EthercatSDO.cpp
@@ -80,7 +80,7 @@ int sdo_bit32_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, u
   int received_working_counter = ec_SDOread(slave, index, sub, FALSE, val_size, value, EC_TIMEOUTRXM);
   if (received_working_counter == 0)
   {
-    ROS_FATAL("(sdo_bit32) Error occurred when writing, reg 0x%X, sub-index %i, data: 0x%X to slave %i", index, sub,
+    ROS_FATAL("(sdo_bit32) Error occurred when reading, reg 0x%X, sub-index %i, data: 0x%X from slave %i", index, sub,
               value, slave);
   }
   return received_working_counter;
--- a/march_hardware/src/EtherCAT/EthercatSDO.cpp
+++ b/march_hardware/src/EtherCAT/EthercatSDO.cpp
@@ -56,7 +56,7 @@ int sdo_bit16_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, u
   int received_working_counter = ec_SDOread(slave, index, sub, FALSE, val_size, value, EC_TIMEOUTRXM);
   if (received_working_counter == 0)
   {
-    ROS_FATAL("(sdo_bit16) Error occurred when writing, reg 0x%X, sub-index %i, data: 0x%X to slave %i", index, sub,
+    ROS_FATAL("(sdo_bit16) Error occurred when reading, reg 0x%X, sub-index %i, data: 0x%X from slave %i", index, sub,
               value, slave);
   }
   return received_working_counter;
--- a/march_hardware/src/EtherCAT/EthercatSDO.cpp
+++ b/march_hardware/src/EtherCAT/EthercatSDO.cpp
@@ -26,9 +26,6 @@ int sdo_bit8_write(uint16_t slave, uint16_t index, uint8_t sub, uint8_t value)
   return received_working_counter;
 }
 
-/**
- * This function returns a working counter. The actual 8 bit read value can be found in the parsed argument; value
- */
 int sdo_bit8_read(uint16_t slave, uint16_t index, uint8_t sub, int& val_size, uint8_t& value)
 {
   ROS_DEBUG("sdo_bit8: slaveIndex %i, reg 0x%X, subindex %i, value 0x%X", slave, index, sub, value);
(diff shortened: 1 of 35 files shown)


march_hardware/src/IMotionCube.cpp
```diff
@@ -82,26 +82,26 @@ void IMotionCube::writeInitialSettings(uint8_t cycle_time)
   ROS_DEBUG("IMotionCube::writeInitialSettings");
 
   // mode of operation
-  int mode_of_op = sdo_bit8(slaveIndex, 0x6060, 0, this->actuation_mode_.toModeNumber());
+  int mode_of_op = sdo_bit8_write(slaveIndex, 0x6060, 0, this->actuation_mode_.toModeNumber());
 
   // position limit -- min position
-  int min_pos_lim = sdo_bit32(slaveIndex, 0x607D, 1, this->absolute_encoder_->getLowerSoftLimitIU());
+  int min_pos_lim = sdo_bit32_write(slaveIndex, 0x607D, 1, this->absolute_encoder_.getLowerSoftLimitIU());
```
Joris-JW:
```suggestion
  int min_pos_lim = sdo_bit32_write(slaveIndex, 0x607D, 1, this->absolute_encoder_->getLowerSoftLimitIU());
```
absolute_encoder_ is a unique pointer (this is also why your build fails

march_hardware/src/IMotionCube.cpp
```diff
@@ -82,26 +82,26 @@ void IMotionCube::writeInitialSettings(uint8_t cycle_time)
...
 
   // position limit -- max position
-  int max_pos_lim = sdo_bit32(slaveIndex, 0x607D, 2, this->absolute_encoder_->getUpperSoftLimitIU());
+  int max_pos_lim = sdo_bit32_write(slaveIndex, 0x607D, 2, this->absolute_encoder_.getUpperSoftLimitIU());
```
Joris-JW:
```suggestion
  int max_pos_lim = sdo_bit32_write(slaveIndex, 0x607D, 2, this->absolute_encoder_->getUpperSoftLimitIU());
```

Joris-JW (dismissed):

march_hardware/src/IMotionCube.cpp
Deakensbubbles: Trying to build this results in base operand of ‘->’ has non-pointer type ‘march::AbsoluteEncoder’ , the other notation does build on my system.

march_hardware/src/EtherCAT/EthercatSDO.cpp
```diff
@@ -26,7 +26,19 @@ int sdo_bit8(uint16_t slave, uint16_t index, uint8_t sub, uint8_t value)
...
 }
 
-int sdo_bit16(uint16_t slave, uint16_t index, uint8_t sub, uint16_t value)
+int sdo_bit8_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, uint8_t* value)
+{
+  ROS_DEBUG("sdo_bit8: slaveIndex %i, reg 0x%X, subindex %i, value 0x%X", slave, index, sub, value);
+  int received_working_counter = ec_SDOread(slave, index, sub, FALSE, val_size, value, EC_TIMEOUTRXM);
+  if (received_working_counter == 0)
+  {
+    ROS_FATAL("(sdo_bit8) Error occurred when writing, reg 0x%X, sub-index %i, data: 0x%X to slave %i", index, sub,
```
Roelemans:
```suggestion
    ROS_FATAL("(sdo_bit8) Error occurred when reading, reg 0x%X, sub-index %i, data: 0x%X from slave %i", index, sub,
```
Maybe de printing the data is also not so useful here :innocent:

march_hardware/src/EtherCAT/EthercatSDO.cpp
```diff
@@ -38,7 +50,19 @@ int sdo_bit16(uint16_t slave, uint16_t index, uint8_t sub, uint16_t value)
...
 }
 
-int sdo_bit32(uint16_t slave, uint16_t index, uint8_t sub, uint32_t value)
+int sdo_bit16_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, uint16_t* value)
+{
+  ROS_DEBUG("sdo_bit16: slaveIndex %i, reg 0x%X, subindex %i, value 0x%X", slave, index, sub, value);
+  int received_working_counter = ec_SDOread(slave, index, sub, FALSE, val_size, value, EC_TIMEOUTRXM);
+  if (received_working_counter == 0)
+  {
+    ROS_FATAL("(sdo_bit16) Error occurred when writing, reg 0x%X, sub-index %i, data: 0x%X to slave %i", index, sub,
```
Roelemans:
```suggestion
    ROS_FATAL("(sdo_bit16) Error occurred when reading, reg 0x%X, sub-index %i, data: 0x%X from slave %i", index, sub,
```

march_hardware/src/EtherCAT/EthercatSDO.cpp
```diff
@@ -26,7 +26,19 @@ int sdo_bit8(uint16_t slave, uint16_t index, uint8_t sub, uint8_t value)
...
+int sdo_bit8_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, uint8_t* value)
+{
+  ROS_DEBUG("sdo_bit8: slaveIndex %i, reg 0x%X, subindex %i, value 0x%X", slave, index, sub, value);
+  int received_working_counter = ec_SDOread(slave, index, sub, FALSE, val_size, value, EC_TIMEOUTRXM);
+  if (received_working_counter == 0)
+  {
+    ROS_FATAL("(sdo_bit8) Error occurred when writing, reg 0x%X, sub-index %i, data: 0x%X to slave %i", index, sub,
+              value, slave);
+  }
+  return received_working_counter;
```
Roelemans, line 38: Shouldn't you return the read value? or is that stored inside the pointer uint8_t* value

Roelemans (changes requested):
Nice small PR! Found a small copy-paste artifact :cat:

march_hardware/src/EtherCAT/EthercatSDO.cpp
RPKrijnen: it is indeed stored inside the pointer. The use of the function is mostly the same as for the write function in this way.

Roelemans (dismissed):
很好

march_hardware/src/EtherCAT/EthercatSDO.cpp
Joris-JW: is it not more useful to convert the function to return a value instead of the value inside the pointer? Calling a function called read and always returning the same variable can cause confusion.

Joris-JW (commented):
just a small question

march_hardware/src/EtherCAT/EthercatSDO.cpp
RPKrijnen: If we do that we miss out on the returned working counter. We're not always returning the same value, this is purely dependent on the type of object that we are reading from. Moreover, it is common practice to produce this form of read and write functions in c code (where you parse a pointer and return a status/error value).

Olavhaasie: In c++ it is more convention to use a reference, instead of a pointer (pointers are easier to get wrong)

Olavhaasie (dismissed):
As I mentioned in my above comment, I think it would be nicer to use references instead of pointers, since we are using C++ and not C like the soem library. References ensure that they have a value and cannot be null. Also you should docstrings above the `sdo_bitx_write` functions to mention that the functions will return its values in the references. An example
```cpp
/**
 * Returns bla...
 */
int sdo_bitx_write(...);
```

march_hardware/src/EtherCAT/EthercatSDO.cpp
RPKrijnen: So to clarify, just do this?:
`int sdo_bitx_read(uint16_t slave, uint16_t index, uint8_t sub, int &val_size, uint8_t &value)`

Olavhaasie: correct, I think that should work

march_hardware/src/EtherCAT/EthercatSDO.cpp
```diff
@@ -26,10 +26,13 @@ int sdo_bit8_write(uint16_t slave, uint16_t index, uint8_t sub, uint8_t value)
   return received_working_counter;
 }
 
-int sdo_bit8_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, uint8_t* value)
+/**
+ * This function returns a working counter. The actual 8 bit read value can be found in the parsed argument; value
```
Roelemans:
```suggestion
 * This method returns a working counter. The actual 8 bit read value can be found in the parsed argument; value
```

march_hardware/src/EtherCAT/EthercatSDO.cpp
```diff
@@ -50,10 +53,13 @@ int sdo_bit16_write(uint16_t slave, uint16_t index, uint8_t sub, uint16_t value)
   return received_working_counter;
 }
 
-int sdo_bit16_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, uint16_t* value)
+/**
+ * This function returns a working counter. The actual 16 bit read value can be found in the parsed argument; value
```
Roelemans:
```suggestion
 * This method returns a working counter. The actual 16 bit read value can be found in the parsed argument; value
```

march_hardware/src/EtherCAT/EthercatSDO.cpp
```diff
@@ -74,10 +80,13 @@ int sdo_bit32_write(uint16_t slave, uint16_t index, uint8_t sub, uint32_t value)
   return received_working_counter;
 }
 
-int sdo_bit32_read(uint16_t slave, uint16_t index, uint8_t sub, int* val_size, uint32_t* value)
+/**
+ * This function returns a working counter. The actual 32 bit read value can be found in the parsed argument; value
```
Roelemans:
```suggestion
 * This method returns a working counter. The actual 32 bit read value can be found in the parsed argument; value
```

Roelemans (commented):
I think you would usually call them methods, not functions. @Olavhaasie wouldn't you normally place these docstrings inside the .h file instead of the cpp file? Please be a bit more specific in your comments :innocent:

march_hardware/src/EtherCAT/EthercatSDO.cpp
RPKrijnen: I always thought it was a conventional difference between Python and c++

RPKrijnen: My mistake, not necessarily a difference between languages, but rather a difference between OOP and regular programming. Because this piece of code is not specific to an object, it is not a method.

Roelemans: O wow I learned something new today

Roelemans (approved):

Joris-JW (approved):
maybe volav wants to take one more look, so I wont merge yet

PR #250, opened 2020-04-03
Check for change in voltage and encoder positions

Roelemans:
## Description

I made the naïve assumption that imc voltage would always change. I forgot that imc voltage, is contained in 16 bits and therefore highly discrete. So instead, we should check if any of imc voltage, or the encoders have updated.

## Changes

* Check whether either voltage or one of the encoders has updated instead of voltage alone.

### DIFF
--- a/march_hardware/include/march_hardware/Joint.h
+++ b/march_hardware/include/march_hardware/Joint.h
@@ -122,6 +122,7 @@ private:
   const int net_number_;
   bool allow_actuation_ = false;
   float previous_imc_volt_ = 0.0;
+  float previous_motor_current_ = 0.0;
   double previous_absolute_position_ = 0.0;
   double previous_incremental_position_ = 0.0;
 
--- a/march_hardware/include/march_hardware/Joint.h
+++ b/march_hardware/include/march_hardware/Joint.h
@@ -122,6 +122,8 @@ private:
   const int net_number_;
   bool allow_actuation_ = false;
   float previous_imc_volt_ = 0.0;
+  double previous_absolute_position = 0.0;
+  double previous_incremental_position = 0.0;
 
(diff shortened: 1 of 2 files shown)


march_hardware/include/march_hardware/Joint.h
```diff
@@ -122,6 +122,8 @@ class Joint
   const int net_number_;
   bool allow_actuation_ = false;
   float previous_imc_volt_ = 0.0;
+  double previous_absolute_position = 0.0;
+  double previous_incremental_position = 0.0;
```
Olavhaasie: :see_no_evil: 
```suggestion
  double previous_absolute_position_ = 0.0;
  double previous_incremental_position_ = 0.0;
```

Olavhaasie (changes requested):

march_hardware/include/march_hardware/Joint.h
Roelemans: Oopsie

Olavhaasie (dismissed):

RutgerVanBeek (changes requested):
You still have a chance (although small) of false positives with this technique. It seems that the current  the most unique values. Therefore I think you should at least include the current in the test. Is there by any chance a PDO mappable object that should be unique be definition, like time, that we could use for this?

Joris-JW (dismissed):
I think the chance of false positives is significantly small.

Roelemans:
The test now also check for current. As Rutger mentioned, it would be ideal if there is a variable in the PDO that always changes, like time. We coincidentally also want to add time to the PDO for encoder data filters, so we will look at that in the future.

RutgerVanBeek (dismissed):

RutgerVanBeek (approved):

Joris-JW (approved):

PR #251, opened 2020-04-06
Hardcoded print all IMC that go to fault state

Ciliaclaij:
First part of PM-336

## Description
Hardcoded all IMC's that go to fault state. After we detect one IMC goes to fault state, we will loop over all the IMC's and print all that are in fault state.

## Changes
Added for loop to detect which IMC's are in fault state


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -448,15 +448,21 @@ void MarchHardwareInterface::iMotionCubeStateCheck(size_t joint_index)
...
+          if (imc_state.state == march::IMCState::FAULT){
+              std::ostringstream error_stream;
+              error_stream << "IMotionCube of joint " << joint_names_[joint_index].c_str() << " is in fault state "
+                           << imc_state.state.getString() << std::endl;
+              error_stream << "Detailed Error: " << imc_state.detailedErrorDescription << "(" << imc_state.detailedError << ")"
+                           << std::endl;
+              error_stream << "Motion Error: " << imc_state.motionErrorDescription << "(" << imc_state.motionError << ")"
+                           << std::endl;
+
+              throw std::runtime_error(error_stream.str());
```
Roelemans: this command throws an error and shuts down the exoskeleton. You should not throw it again for every imc that is in fault state, but rather only once at the end if any imc has faulted. 

You can construct the error message by declaring `error_stream` before the loop, and then adding text on every iteration.

Roelemans (changes requested):

march_hardware_interface/src/march_hardware_interface.cpp
Ciliaclaij: So `std::ostringstream error_stream;` needs to be before the loop and `throw std::runtime_error(error_stream.str());` after the loop?

Joris-JW: I would change the std::runtime_error with a logging statement and place the throw after the loop.

Joris-JW (dismissed):
Throw error after the loop

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -448,15 +448,23 @@ void MarchHardwareInterface::iMotionCubeStateCheck(size_t joint_index)
...
-    std::ostringstream error_stream;
-    error_stream << "IMotionCube of joint " << joint_names_[joint_index].c_str() << " is in fault state "
-                 << imc_state.state.getString() << std::endl;
-    error_stream << "Detailed Error: " << imc_state.detailedErrorDescription << "(" << imc_state.detailedError << ")"
-                 << std::endl;
-    error_stream << "Motion Error: " << imc_state.motionErrorDescription << "(" << imc_state.motionError << ")"
-                 << std::endl;
-
-    throw std::runtime_error(error_stream.str());
+    for (int i = 0; i < 8; i++)
```
Olavhaasie: This `i` variable is never used in this for loop. Instead, this method is called for each of the joints, which is indicated by the `joint_index` parameter. You could make the method return a `bool` on whether or not the state check was successful and log if there was a motion error. Then in the `validate` function you can check for all joints that fail and then throw an exception.

Olavhaasie (changes requested):

march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
```diff
@@ -130,6 +130,7 @@ class MarchHardwareInterface : public hardware_interface::RobotHW
   bool master_shutdown_allowed_command_ = false;
   bool enable_high_voltage_command_ = true;
   bool reset_imc_ = false;
+  bool fault_state = false;
```
Roelemans:
```suggestion
```
Instead of using a member variable like this, it is easier to return a bool in the iMotionCubeStateCheck method :)

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -191,6 +191,11 @@ void MarchHardwareInterface::validate()
     this->outsideLimitsCheck(i);
     this->iMotionCubeStateCheck(i);
```
Roelemans:
```suggestion
    if (this->iMotionCubeStateCheck(i)) {
      fault_state = true;
    }
```
Make sure to also declare `fault_state` before the loop (after line 188)

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -444,20 +449,18 @@ void MarchHardwareInterface::updateIMotionCubeState()
...
+    march::IMotionCubeState imc_state = march_robot_->getJoint(joint_names_[joint_index]).getIMotionCubeState();
+    if (imc_state.state == march::IMCState::FAULT)
+    {
+        ROS_ERROR(<< "IMotionCube of joint " << joint_names_[joint_index].c_str() << " is in fault state "
+                  << imc_state.state.getString() << std::endl );
+        ROS_ERROR(<< "Detailed Error: " << imc_state.detailedErrorDescription << "(" << imc_state.detailedError << ")"
+                  << std::endl);
+        ROS_ERROR(<< "Motion Error: " << imc_state.motionErrorDescription << "(" << imc_state.motionError << ")"
+                  << std::endl;)
+        fault_state = true;
```
Roelemans:
```suggestion
        return false;
```
(it makes more sense to return false when something goes wrong and true when everything is okay)

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -444,20 +449,18 @@ void MarchHardwareInterface::updateIMotionCubeState()
...
+
+    }
 }
```
Roelemans, line 456:
```suggestion
  return true;
}
```

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -444,20 +449,18 @@ void MarchHardwareInterface::updateIMotionCubeState()
 
 void MarchHardwareInterface::iMotionCubeStateCheck(size_t joint_index)
```
Roelemans:
```suggestion
bool MarchHardwareInterface::iMotionCubeStateCheck(size_t joint_index)
```
Also make this change in the header file

Roelemans (changes requested):
Good job, this looks really good :) one small suggestion for improvement and a tip to fix your build:

Your clang format check fails (this a style chek) you should run
`find . -name '*.h' -or -name '*.cpp' | xargs clang-format -i -style=file`
This is a long command that noone ever remembers, so it is advisable to add a shortcut to your .bashrc file. i.e. open the .bashrc file in your home directory and add to the bottom of the file:
`alias format_clang="find . -name '*.h' -or -name '*.cpp' | xargs clang-format -i -style=file"`

march_hardware_interface/src/march_hardware_interface.cpp
Olavhaasie: Down here, you say that it should return `false` when something is wrong and `true` when everything's good, so you should invert this logic statement

Ciliaclaij: I don't understand the declaration of the variable after line 188. I thought variables need to be declared in the header file?

Olavhaasie: variables declared in the class header are considered member variables and can be accessed through the `this` object. Local variables, like the one Roel suggests can be declared inside a function and will only be accessible inside the function. Since you only use this variable inside this function, it is better to declare it after line 188.

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -186,10 +186,19 @@ bool MarchHardwareInterface::init(ros::NodeHandle& nh, ros::NodeHandle& /* robot
 
 void MarchHardwareInterface::validate()
 {
+  bool fault_state;
```
Olavhaasie: It is better to initialize your variables with a value, then your intent is more clear, i.e. default is no fault state.
```suggestion
  bool fault_state = false;
```

Olavhaasie (changes requested):

Roelemans (dismissed):
Good work! :)

Olavhaasie (dismissed):

Joris-JW (commented):
Nice, proper clean solution.

Joris-JW (dismissed):

RutgerVanBeek (changes requested):
It seems like have you some conversion errors when printing. See Slack for the exact ones.

RutgerVanBeek (dismissed):
I think this is indeed the fix. Should check on exo next week (Or convince electro to do it this afternoon)

Joris-JW:
Has this been successfully tested on exo? in that case the label can be removed.

Olavhaasie (approved):

RutgerVanBeek:
> Has this been successfully tested on exo? in that case the label can be removed.

Last week, it worked except for the printing.

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -430,23 +439,20 @@ void MarchHardwareInterface::updateIMotionCubeState()
...
-    error_stream << "IMotionCube of joint " << joint.getName() << " is in fault state " << imc_state.state.getString()
-                 << std::endl;
-    error_stream << "Detailed Error: " << imc_state.detailedErrorDescription << "(" << imc_state.detailedError << ")"
-                 << std::endl;
-    error_stream << "Motion Error: " << imc_state.motionErrorDescription << "(" << imc_state.motionError << ")"
-                 << std::endl;
-
-    throw std::runtime_error(error_stream.str());
+    ROS_ERROR("IMotionCube of joint %d is in fault state %s \n", joint.getName().c_str(),
+              imc_state.state.getString().c_str());
```
Olavhaasie: Found a tiny teensy oopsie. `%d` will format for integers, however, the first parameter is a string.
```suggestion
    ROS_ERROR("IMotionCube of joint %s is in fault state %s \n", joint.getName().c_str(),
              imc_state.state.getString().c_str());
```

Olavhaasie (changes requested):

Olavhaasie (approved):

RutgerVanBeek (approved):

PR #252, opened 2020-04-06
Feature: URDF joint consistency check

Olavhaasie:
Closes PM-376

## Description
In this PR I used the URDF to find inconsistencies between the URDF joint and the robot joints. I added 3:

* When a joint exists in the URDF and is not fixed and is not in the robots yaml, then log a warning
* When a joint exists in the URDF is fixed and is in the robots yaml, then log a warning
* When a joint exists in the robots yaml and does not have a URDF joint, then throw an exception

This exception is the same as the exceptions thrown in `march_hardware`. I added the error types to the list in `march_hardware`. I know these errors don't occur in the `march_hardware`, but this was the best way to avoid writing the same code in both packages. We might have to create a separate library, if we need it at more places in our codebase.

Furthermore, I removed adding the non-fixed joint names from the URDF in the hardware interface. I did this because the joints that need to be controlled are in the march robot and not the URDF. This also removes the need to make joints fixed in the URDF, when they should not be controlled. The order of the joints in the hardware interface now resemble the order in the robots yaml and not the URDF. However, in the URDF they are sorted by name, which is not the case for the robots yaml. I don't think this an issue, but if it is then we have to sort that file as well. I have already made sure to sort the joint names before uploading them to the parameter server. While I was doing this, I found that iterating over the joints was a bit inefficient, since it was finding them by name every loop. I optimized this by adding getter on the `MarchRobot` by joint index and I implemented a iterator interface, so now you can use a for each loop. The reason I could not use the for each loop on every loop is that those loops require the index for the controller variables.

## Changes
* Added  consistency checks for URDF and robot joints
* Add error types to `march_hardware_builder`
* Added catching exceptions from the hardware builder in `march_hardware_interace_node`
* Add tests for new checks
* Add iterator interface on `MarchRobot` over joints
* Remove adding joints by name from URDF in `MarchHardwareInterface`
* Use new indexing of joints in `MarchHardwareInterface`
* Sort joint names before uploading


RutgerVanBeek:
See Slack for the error we got during testing.

Roelemans:
Maybe close this PR until a solution is found? @Olavhaasie

Joris-JW (dismissed):
I think your last commit indeed solved the, weird, error. But let's wait for the exo-test just in case.

march_hardware/src/error/error_type.cpp
```diff
@@ -43,6 +43,12 @@ const char* getErrorDescription(ErrorType type)
       return "A joint is not allowed to actuate, yet it's trying to actuate";
     case ErrorType::INVALID_SLAVE_INDEX:
       return "Slave index has an invalid value";
+    case ErrorType::MISSING_URDF_JOINT:
+      return "Required joint not defined in URDF";
+    case ErrorType::MISSING_REQUIRED_KEY:
+      return "Required robot config key not defined";
+    case ErrorType::INIT_URDF_FAILED:
+      return "Failed to initialize URDF";
```
Roelemans:
```suggestion
      return "Failed to load URDF from parameter server";
```
I think this is more descriptive of what went wrong?

march_hardware_builder/include/march_hardware_builder/hardware_config_exceptions.h
```diff
@@ -1,42 +1,21 @@
...
-#include <exception>
+#include <sstream>
 #include <string>
 
-#include <ros/ros.h>
+#include <march_hardware/error/error_type.h>
+#include <march_hardware/error/hardware_exception.h>
 
-class HardwareConfigException : public std::exception
+class MissingKeyException : public march::error::HardwareException
```
Roelemans, line 10: Why is it necessary to have a separate class for this if there is also a separate type inside the HardwareException?

march_hardware/include/march_hardware/MarchRobot.h
```diff
@@ -58,6 +61,16 @@ class MarchRobot
...
 
+  Joint& getJoint(size_t index);
+
+  size_t size() const;
+
+  iterator begin();
+  iterator end();
+
+  const_iterator begin() const;
+  const_iterator end() const;
```
Roelemans: Are there locations in the code where there is a const MarchRobot?

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -260,12 +257,43 @@ void HardwareBuilder::initUrdf()
...
+  std::vector<march::Joint> joints;
+  for (const YAML::Node& joint_config : joints_config)
+  {
+    const auto joint_name = joint_config.begin()->first.as<std::string>();
+    const auto urdf_joint = this->urdf_.getJoint(joint_name);
+    if (urdf_joint->type == urdf::Joint::FIXED)
+    {
+      ROS_WARN("Joint %s is fixed in the URDF, but defined in the robot yaml", joint_name.c_str());
+    }
+    joints.push_back(HardwareBuilder::createJoint(joint_config[joint_name], joint_name, urdf_joint));
```
Roelemans, line 277: If the joint is fixed in the URDF, you still add it to the joint list? Won't this cause problems if the soft limits for this joint are not defined in the URDF?

Roelemans (changes requested):
Nice changes :)

march_hardware_builder/include/march_hardware_builder/hardware_config_exceptions.h
Olavhaasie: So that the name of the missing key for an object can be used to construct the message, since that is not possible in `HardwareException`

march_hardware/include/march_hardware/MarchRobot.h
Olavhaasie: The call doesn't have to be from a `const MarchRobot`. The `const` on the end means that the object is not modified during this function call and the `const_iterator` means the iterator iterates over `const Joint`, which do not allow you to modify them. For example, this is used when you want to loop over all joints to get their names (which is also constant). Furthermore, this allows the compiler to do some optimizations.

march_hardware_builder/src/hardware_builder.cpp
Olavhaasie: The warning is here because the joint was defined in the robots yaml, but is FIXED in the URDF. It is possible to continue. If there are no soft limits it will warn you when trying to construct the actual joint.

march_hardware/include/march_hardware/MarchRobot.h
Roelemans: But there exists a method begin() and a const method begin(). According to my limited understanding, the non-const method wil get called unless it is called from a const MarchRobot.

Roelemans (approved):

Roelemans (changes requested):
Approve oopsie should have been comment or request changes

march_hardware/include/march_hardware/MarchRobot.h
Olavhaasie: I thought the `const_iterator` would be used for
```cpp
for (const auto& joint : march_robot)
{
...
}
```
But apparently it turns out [it doesn't](https://stackoverflow.com/questions/26554636/why-doesnt-const-range-based-for-use-const-iterator). However, what you are saying is not completely true, since you can use the const_iterator like so:
```cpp
march::MarchRobot robot(...);  // notice it is not const
march::MarchRobot::const_iterator it = robot.begin();
```
This is not the best use case since you should always use `for` loops. However, when you define this method you can also loop over joints when `MarchRobot` _is_ defined const. But, as you mentioned, we don't do that anywhere. So we could leave it in for the future, I think it could be useful for writing tests, where you create a robot and then only read, but it is definitely not necessary.

Roelemans: In that case, I would remove it as it is not currently in use and has a small chance of being used (outside tests) in the future. At that point, one can always make these methods again.

Roelemans (changes requested):

Roelemans (approved):
:)

march_hardware/include/march_hardware/MarchRobot.h
```diff
@@ -58,6 +60,13 @@ class MarchRobot
 
   Joint& getJoint(::std::string jointName);
 
+  Joint& getJoint(size_t index);
+
+  size_t size() const;
+
+  iterator begin();
+  iterator end();
```
Joris-JW, line 68: I would go for a more descriptive name if not to much trouble, begin and end could also refer to something like the initiating of the robot.
```suggestion
  size_t JointListSize() const;

  iterator JointListBegin();
  iterator JointListEnd();
```

Joris-JW (approved):
got a small comment but u can decide if u want to honor it.  code looks good!

march_hardware/include/march_hardware/MarchRobot.h
Olavhaasie: the `begin` and `end` functions have to have that name otherwise the for each loop won't work. I have named the other `size` to be consistent with `std::vector`.

PR #253, opened 2020-04-13
Feature/pm 382 write checksum method

RPKrijnen:
Closes PM-382

## Description

This PR implements functionality to compute the checksum on the IMC as well as on the intended setup of that IMC.

## Changes

*Added new constructor that accepts one more argument
*Added new private variable containing entire setup file of the corresponding imc
*Added functionality to read out .sw file and send it through in form of a stringstream
*Added a function that righthand-justifies a string of fewer than four numbers (adds more zeros to the left).


march_hardware_builder/robots/test_joint_rotational.yaml
```diff
@@ -1,5 +1,5 @@
 testsetup:
-  ifName: enp2s0
```
Joris-JW: This should not be changed, the name corresponds with the ip-link port name on the exoskeleton.

RPKrijnen: Well spotted, my mistake :)

Joris-JW: that was fast.

RPKrijnen: You think I have anything better to do? haha

march_hardware/src/IMotionCube.cpp
```diff
@@ -21,12 +28,36 @@ IMotionCube::IMotionCube(int slave_index, std::unique_ptr<AbsoluteEncoder> absol
...
+  , absolute_encoder_(std::move(absolute_encoder))
+  , incremental_encoder_(std::move(incremental_encoder))
+  , sw_stream_(std::move(sw_stream))
+  , actuation_mode_(actuation_mode)
+{
+  if (!this->absolute_encoder_ || !this->incremental_encoder_)
+  {
+    throw std::invalid_argument("Incremental or absolute encoder cannot be nullptr");
+  }
+  std::cout << "length in constructor: " << this->sw_stream_.str().length() << std::endl;
```
Joris-JW: This logging statement is twice in the constructor, also at line 38. Isn't the first statement redundant since no value has been assigned to it except the "empty"?

RPKrijnen: This was a copy paste error, I shall rectify it as it doesn't serve any purpose to print the length of the .sw file in the default constructor.

march_hardware/src/IMotionCube.cpp
```diff
@@ -331,6 +361,33 @@ void IMotionCube::reset()
...
 }
 
+uint16_t IMotionCube::computeSWCheckSum(int& start_address, int& end_address)
+{
+  size_t pos = 0;
+  size_t old_pos = 0;
+  uint16_t sum = 0;
+  std::string delimiter = "\n";
+  std::string token;
+  while ((pos = sw_stream_.str().find(delimiter, old_pos)) != std::string::npos)
```
Olavhaasie: There is no use in using a `std::stringstream` if the first thing you are going to do is reading out the entire stream. Streams are there for data that is not yet complete, for example, reading from files. Therefore, if you need the complete string in order to compute the checksum, you shouldn't be using streams, but strings. I think the implementation would become much lighter if you use a `std::string`

march_hardware/src/IMotionCube.cpp
```diff
@@ -35,6 +43,28 @@ IMotionCube::IMotionCube(int slave_index, std::unique_ptr<AbsoluteEncoder> absol
...
+  }
+  ROS_INFO("length in constructor: %d", this->sw_stream_.str().length());
+  this->absolute_encoder_->setSlaveIndex(slave_index);
+  this->incremental_encoder_->setSlaveIndex(slave_index);
+  this->is_incremental_more_precise_ =
+      (this->incremental_encoder_->getTotalPositions() * this->incremental_encoder_->getTransmission() >
+       this->absolute_encoder_->getTotalPositions() * 10);
+  // Multiply by ten to ensure the rotational joints keep using absolute encoders. These are somehow more accurate
+  // even though they theoretically shouldn't be.
+}
```
Olavhaasie: To avoid code redundancy, you can call the other constructor from the initializer list.
```suggestion
  : IMotionCube(slave_index, std::move(absolute_encoder), std::move(incremental_encoder), actuation_mode)
{
  this->sw_stream_ = std::move(sw_stream);
  ROS_INFO("length in constructor: %lu", this->sw_stream_.str().length());
}
```

march_hardware/test/TestIMotionCube.cpp
```diff
@@ -8,6 +8,7 @@
 #include <memory>
 #include <stdexcept>
 #include <utility>
+#include <sstream>
```
Olavhaasie: Please remove unused imports

march_hardware/test/mocks/MockIMotionCube.cpp
```diff
@@ -6,6 +6,7 @@
 #include <memory>
 
 #include <gmock/gmock.h>
+#include <sstream>
```
Olavhaasie: Same here

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -129,15 +133,23 @@ std::unique_ptr<march::IMotionCube> HardwareBuilder::createIMotionCube(const YAM
   {
     return nullptr;
   }
-
+  // std::cout << imc_config;
```
Joris-JW: I think this can be removed

march_hardware/include/march_hardware/IMotionCube.h
```diff
@@ -96,13 +100,15 @@ class IMotionCube : public Slave
   void mapMisoPDOs();
   void mapMosiPDOs();
   void writeInitialSettings(uint8_t cycle_time);
+  uint16_t computeSWCheckSum(int& start_address, int& end_address);
```
Olavhaasie: I think it would be useful if you add documentation on this method to describe what it returns and what is expected from the two int reference parameters.

march_hardware_builder/src/hardware_builder.cpp
Olavhaasie: Please remove this comment :wink:

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -257,3 +269,24 @@ void HardwareBuilder::initUrdf()
...
+  std::string delimiter = "\n";
+  if (input.length() > delimiter.length())
+  {
+    while (input.size() < 4 + delimiter.length())  // check until the 16-bit number has been filled
+    {
+      input.insert(0, "0");
+    }
+  }
+  return input;
+}
```
Olavhaasie: Is there a reason that is method is not being used?

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -257,3 +269,24 @@ void HardwareBuilder::initUrdf()
...
+
+std::stringstream convertSWFileToStringStream(std::ifstream& sw_file)
+{
+  std::stringstream sw_stream;
+  std::copy(std::istreambuf_iterator<char>(sw_file), std::istreambuf_iterator<char>(),
+            std::ostreambuf_iterator<char>(sw_stream));
+  return sw_stream;
+}
```
Olavhaasie: See my above comment about using a string instead of a stream. You could change this method like so
```suggestion
std::string convertSWFileToString(std::ifstream& sw_file)
{
  return std::string(std::istreambuf_iterator<char>(sw_file),
                     std::istreambuf_iterator<char>());
}
```

march_hardware_builder/include/march_hardware_builder/hardware_builder.h
```diff
@@ -93,4 +93,13 @@ class HardwareBuilder
   bool init_urdf_ = true;
 };
 
+/*
```
Olavhaasie: docstrings should start with `/**` :innocent: 
```suggestion
/**
```

march_hardware_builder/include/march_hardware_builder/hardware_builder.h
```diff
@@ -93,4 +93,13 @@ class HardwareBuilder
...
+ * Adds zeros to left of unsigned int (which is in the form of a string) until the 16 bits have been accounted for
+ */
+std::string rightHandJustifyString(std::string input);
+/*
```
Olavhaasie:
```suggestion
/**
```

march_hardware_builder/include/march_hardware_builder/hardware_builder.h
```diff
@@ -93,4 +93,13 @@ class HardwareBuilder
   bool init_urdf_ = true;
 };
 
+/*
+ * Adds zeros to left of unsigned int (which is in the form of a string) until the 16 bits have been accounted for
+ */
+std::string rightHandJustifyString(std::string input);
```
Joris-JW: I cannot find the usage of this function in this PR, am I missing something?

Olavhaasie (changes requested):

march_hardware_builder/src/hardware_builder.cpp
Joris-JW: @Olavhaasie already commented this so i will resolve mine!

march_hardware_builder/include/march_hardware_builder/hardware_builder.h
Joris-JW: @Olavhaasie already commented this so i will resolve mine!

march_hardware_builder/src/hardware_builder.cpp
RPKrijnen: I have previously used an iterating function that made use of this one. The current implementation used a type of copying to move from ifstream to stringstream. I forgot to add this function to the converter.

march_hardware/src/IMotionCube.cpp
RPKrijnen: could you please elaborate, either here or via Slack. I do not fully understand what you want changed even though you have provided a suggestion.

march_hardware_builder/src/hardware_builder.cpp
RPKrijnen: I have adjusted the code so that it is used again, but the use case for the function has expanded to strings containing all the data of the file.

march_hardware/include/march_hardware/IMotionCube.h
RPKrijnen: I will provide a docstring above the function for clarification.

march_hardware/include/march_hardware/IMotionCube.h
```diff
@@ -96,13 +99,23 @@ class IMotionCube : public Slave
...
+   */
+  uint32_t computeSWCheckSum(int& start_address, int& end_address);
 
   // Use of smart pointers are necessary here to make dependency injection
   // possible and thus allow for mocking the encoders. A unique pointer is
   // chosen since the IMotionCube should be the owner and the encoders
   // do not need to be passed around.
   std::unique_ptr<AbsoluteEncoder> absolute_encoder_;
   std::unique_ptr<IncrementalEncoder> incremental_encoder_;
+  std::string sw_stream_;
```
Olavhaasie: Perhaps rename this variable since it is no longer a stream? Seems a bit confusing when you see the name referenced somewhere else in the code.

march_hardware/src/IMotionCube.cpp
```diff
@@ -35,6 +40,14 @@ IMotionCube::IMotionCube(int slave_index, std::unique_ptr<AbsoluteEncoder> absol
   // Multiply by ten to ensure the rotational joints keep using absolute encoders. These are somehow more accurate
   // even though they theoretically shouldn't be.
 }
+IMotionCube::IMotionCube(int slave_index, std::unique_ptr<AbsoluteEncoder> absolute_encoder,
+                         std::unique_ptr<IncrementalEncoder> incremental_encoder, std::string& sw_stream,
+                         ActuationMode actuation_mode)
+  : IMotionCube(slave_index, std::move(absolute_encoder), std::move(incremental_encoder), actuation_mode)
+{
+  this->sw_stream_ = std::move(sw_stream);
+  ROS_INFO("length in constructor: %lu", this->sw_stream_.length());
```
Olavhaasie: Now that I think about it, this info message isn't very descriptive and is probably only necessary for debugging. Therefore, you should either change it to `ROS_DEBUG` and add some information, like slave index, or remove the statement altogether.

march_hardware/src/IMotionCube.cpp
```diff
@@ -331,6 +344,33 @@ void IMotionCube::reset()
...
+    if (pos - old_pos < 2)  // delimiter has length of 1 two \n in a row has difference in positions of 1
+    {
+      end_address = end_address + start_address - 2;  // The -2 compensates the offset of the end_address
+      return sum;
+    }
+    end_address++;
+    sum += std::stoi(token, nullptr, 16);  // State that we are looking at hexadecimal numbers
+    old_pos = pos + 1;                     // Make sure to check the position after the previous one
+  }
+  return sum;  // Only reached if something goes wrong and the code doesn't enter the while loop
```
Olavhaasie: Perhaps you should throw an exception here instead, otherwise it would be impossible to know for the caller that the method failed. You can use the `HardwareException` and create a custom `ErrorType` with it. See `error/error_types.h` and `error/error_types.cpp` for examples.

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -257,3 +270,30 @@ void HardwareBuilder::initUrdf()
...
+      while (token.size() < 4 + delimiter.length())  // check until the 16-bit number has been filled
+      {
+        token.insert(0, "0");
+      }
+    }
+    result += input;
+    old_pos = pos + 1;  // Make sure to check the position after the previous one
+  }
+  return input;
+}
```
Olavhaasie: Why is this method even necessary in the first place? Wouldn't the values `00AB` and `AB` be the same according to `std::stoi`? Furthermore, [`substr`](https://www.cplusplus.com/reference/string/string/substr/) returns a newly constructed string, so the `"0"` s you are inserting are discarded.

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -135,9 +139,18 @@ std::unique_ptr<march::IMotionCube> HardwareBuilder::createIMotionCube(const YAM
   YAML::Node incremental_encoder_config = imc_config["incrementalEncoder"];
   YAML::Node absolute_encoder_config = imc_config["absoluteEncoder"];
   int slave_index = imc_config["slaveIndex"].as<int>();
+
+  std::ifstream imc_setup_data;
+  imc_setup_data.open(ros::package::getPath("march_hardware_builder").append("/config/" + urdf_joint->name + ".sw"));
+  std::cout << "jointname: " << urdf_joint->name << "path to file"
+            << ros::package::getPath("march_hardware_builder").append("/config/" + urdf_joint->name + ".sw")
+            << std::endl;
```
Olavhaasie: Please remove these debug `cout` statements or replace them with the ROS log macros.
```suggestion
```

Olavhaasie (changes requested):

march_hardware_builder/src/hardware_builder.cpp
RPKrijnen: You are right, it is no longer necessary. In previous versions it was useful for finding the previous token in the string because this made the string more predictable, but I have since moved on from that method and forgot to remove this function.

march_hardware/include/march_hardware/IMotionCube.h
RPKrijnen: good suggestion, this is implemented in the last commits

Olavhaasie (dismissed):

Joris-JW (dismissed):

Joris-JW (approved):

Olavhaasie (approved):

PR #255, opened 2020-04-15
Feature/pm 327 use incremental encoders for position

Roelemans:
Closes PM-327

## Description

Use incremental encoders for position. Move speed and position calculation for MarchHardwareInterface logic to Joint.

## Changes

* Use most accurate (the incremental) encoder for position calculation. 
* Move speed and position calculation logic to Joint.
* Write a lot of tests for the Joint class


march_hardware/src/Joint.cpp
```diff
@@ -77,34 +83,65 @@ void Joint::actuateRad(double target_position)
...
-    ROS_WARN("[%s] Has no iMotionCube", this->name_.c_str());
-    return -1;
+    const double new_incremental_position = this->imc_->getAngleRadIncremental();
+    const double new_absolute_position = this->imc_->getAngleRadAbsolute();
+
+    // Get velocity from encoder position
+    double best_displacement;
+
+    // Take the velocity with the highest resolution.
+    if (this->imc_->getIncrementalRadPerBit() < this->imc_->getAbsoluteRadPerBit())
```
RutgerVanBeek, line 103: I wouldn't evaluate this expression every time. I think you should make a boolean per joint. Can you redirect function calls in c++ like you can in Python?

RutgerVanBeek (changes requested):

march_hardware/src/Joint.cpp
Roelemans: I thought this was neater, because it requires less code while the evaluation of the above expression is very close to zero.

Roelemans: I thought this was neater, because it requires less code while the evaluation of the above expression is very close to zero.

RutgerVanBeek: But it is a constant property of a joint that either the incremental or the absolute encoder is the most precise. It is not a big deal, but may be someone else can make the call on what should be used. I realize now you can of course not reassign a function at runtime, because it is compiled.

Olavhaasie: You could use [`std::function`](https://en.cppreference.com/w/cpp/utility/functional) for this, but I think that would make things unnecessarily complicated. You could also just store a `bool` on which encoder has a higher resolution, saves you 2 divide operations and a float comparison. It's up to you :shrug:

RutgerVanBeek (approved):

march_hardware/src/Joint.cpp
Roelemans: Imma leave it like this. I think it is easier to follow exactly when the condition is true this way, because you don't have to look in multiple places to find all the logic. 2 divide operations and a float comparison aren't a very big resource sink :)

Joris-JW: The ```bool``` suggestions seems nice. But since it's up to you I will approve the branch but not merge.

Joris-JW (approved):

PR #256, opened 2020-04-15
Feature/pm 300 add motor voltage

Ciliaclaij:
Closes PM-300

## Description

Add motor voltage reading and publishing on imc state topic

## Changes

e.g.
* Added function to read the motor voltage from the imc
* Publish motor voltage on imc state topic


march_hardware/src/IMotionCube.cpp
```diff
@@ -266,6 +266,15 @@ float IMotionCube::getIMCVoltage()
          static_cast<float>(imc_voltage_iu);  // Conversion to Volt, see Technosoft CoE programming manual
 }
 
+float IMotionCube::getMotorVoltage()
+{
+  uint16_t motor_voltage = 0;
+  int val_size = 16;
+  sdo_bit32_write(slaveIndex, 0x2064, 0, 0x0232);
+  sdo_bit16_read(slaveIndex, 0x2066, 0, val_size, motor_voltage);
```
Olavhaasie: Is it possible to read and write to SDOs, while ethercat is fully operational? I thought they were one-time messages, only sent at initialization. And since they are non-cyclic, won't it take a whole ethercat cycle to read out this SDO?

Olavhaasie (changes requested):

Joris-JW (dismissed):
I'm not sure whether sending and receiving an SDO is sufficient enough to execute every PDO cycle. This is something we have to ensure by testing thoroughly. Besides this, nice small PR!

march_hardware/src/IMotionCube.cpp
Ciliaclaij: I am not sure, @Deakensbubbles requested to use this method in one of my previous pull request so I thought it was fine to use this. But @Roelemans and @JorisWeeda are also concerned about this, so maybe we need to find another solution. But I don't know what to do, so maybe nice to discuss this with some people?

Joris-JW: I've taken the time to read a little bit more about the response time of an SDO request. I do think it is possible to request a value using this structure but the receive timeout stated in SOEM is 0.7 seconds for the SDO whereas our PDO timing is significantly less. I think a much more interesting problem is that the received value does not have a reliable timestamp because the delay can differ. Resulting in a value which may not be usable.

Olavhaasie:
@Ciliaclaij Should we close this PR until you have found another way of reading out motor voltage?

Olavhaasie:
I am closing this PR due to inactivity, feel free to open it once you have new information

PR #257, opened 2020-04-22
Feature/pm 383 write downoad to drive method for sw files

RPKrijnen:
Closes PM-383

## Description

This PR implements methods to download the .sw files to the IMotionCubes. These files are downloaded only when the checksums of the drive and the fie itself are different. After downloading the files the system resets. 
## Changes

* Added methods in IMotionCube.cpp for downoading the files
* fixed return types of functions throughout the HWI to have the MarchRobot restart when necessary.


march_hardware/include/march_hardware/IMotionCube.h
```diff
@@ -99,12 +99,20 @@ class IMotionCube : public Slave
...
+  bool writeInitialSettings(uint8_t cycle_time);
   /**
    * Calculates checksum on .sw file passed in string format in sw_string_ by simple summation until next empty line.
    * Start_address and end_address are filled in the method and used for downloading the .sw file to the drive.
    */
   uint32_t computeSWCheckSum(int& start_address, int& end_address);
+  /**
+   * Compares the checksum of the .sw file and the setup on the drive. If both are equal 1 is returned.
+   */
+  int verifySetup();
```
RutgerVanBeek: Why not a boolean?

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -49,10 +49,11 @@ void EthercatMaster::waitForPdo()
   this->pdo_received_ = false;
 }
 
-void EthercatMaster::start(std::vector<Joint>& joints)
+bool EthercatMaster::start(std::vector<Joint>& joints)
 {
   this->ethercatMasterInitiation();
-  this->ethercatSlaveInitiation(joints);
+  bool reset = this->ethercatSlaveInitiation(joints);
+  return reset;
```
RutgerVanBeek:
```suggestion
  return this->ethercatSlaveInitiation(joints);
```

march_hardware/src/IMotionCube.cpp
```diff
@@ -84,9 +84,19 @@ void IMotionCube::mapMosiPDOs()
...
 
 // Set configuration parameters to the IMC
-void IMotionCube::writeInitialSettings(uint8_t cycle_time)
+bool IMotionCube::writeInitialSettings(uint8_t cycle_time)
 {
-  ROS_DEBUG("IMotionCube::writeInitialSettings");
+  int checksum_verified = verifySetup();
+  ROS_INFO("The .sw file for slave %d is %s", this->slaveIndex,
+           checksum_verified ? "equal to the setup of the drive." :
+                               "not equal to the setup of the drive, downloading is necessary");
```
RutgerVanBeek: I think them not being equal should be considered a special event. Therefore I suggest make it a debug when successful and a warning when unsuccessful. It should not be necessary in normal use.

march_hardware/src/IMotionCube.cpp
```diff
@@ -84,9 +84,19 @@ void IMotionCube::mapMosiPDOs()
...
+  if (!checksum_verified)
+  {
+    downloadSetupToDrive();
+    checksum_verified = verifySetup();
+    ROS_INFO(checksum_verified ? "writing of the setup data has succeeded" : "writing of the setup data has failed");
```
RutgerVanBeek: Isn't this more severe, may be even fatal. Why know there is a wrong iMC project and we are not able to recover. Should we really continue?

march_hardware/src/IMotionCube.cpp
```diff
@@ -364,6 +376,91 @@ uint32_t IMotionCube::computeSWCheckSum(int& start_address, int& end_address)
...
+
+  ROS_DEBUG("The .sw checksum is : %d, and the drive checksum is %d", sw_value, imc_value);
+  if (sw_value == imc_value)
+  {
+    return 1;
+  }
+  else
+  {
+    return 0;
+  }
```
RutgerVanBeek: Why not make this function boolean and do `return sw_value==imc_value` ?

march_hardware/src/MarchRobot.cpp
```diff
@@ -34,28 +34,40 @@ MarchRobot::MarchRobot(::std::vector<Joint> jointList, urdf::Model urdf, PowerDi
 
 void MarchRobot::startEtherCAT(bool reset_imc)
 {
-  if (!hasValidSlaves())
+  bool sw_reset = false;
+  for (int i = 0; i < 5; i++)
```
RutgerVanBeek: Where does the five come from?

RutgerVanBeek (dismissed):
This is really promising I think! I have some small comments/questions. 

Furthermore review requests are also assigned automatically. There is no need to ask for reviews unless you want a specific person to have a look.

march_hardware/include/march_hardware/IMotionCube.h
RPKrijnen: I wanted to use the return of this function just like the SDO commands as a means to verify if the code has succesfully written everything away. The return values of those functions is an int and I thought it would only make sense to do the same for my method.

march_hardware/src/IMotionCube.cpp
RPKrijnen: see my other comment on this

march_hardware/src/MarchRobot.cpp
RPKrijnen: It was arbitrary. We should at least try a few times to recover the system, once might just not be enough.

march_hardware/src/IMotionCube.cpp
RPKrijnen: I want the system to be able to recover from a potential msitake. Making it fatal would kill the node and I want to have a robust way of downloading these files. We have encountered many experts who suggest to just try again and see if the same thing happens before you jump to conclusions after a single failure.

march_hardware/src/IMotionCube.cpp
RPKrijnen: This is a good suggestion, I will implement this shortly.

march_hardware/src/MarchRobot.cpp
RutgerVanBeek: It feels a bit like restarting iMC' s until they work. Anyone else any thoughts?

march_hardware/include/march_hardware/IMotionCube.h
Roelemans: I believe those methods return a working counter? I think it would be neater to either do that too (by returning the outcome of IMotionCube.cpp, line 448) or return a bool

march_hardware/src/IMotionCube.cpp
Roelemans: From what I saw on Wednesday, this does not really fail (often). I think you can just make this a fatal and remove the loop at in MarchRobot.cpp (so only try once). If it fatals very rarely, we can simply launch again. If this fatals too often we can always implement the loop again.

march_hardware/src/MarchRobot.cpp
Roelemans: I think this is not necessary since we have not seen this fail even once last Wednesday. Hence, I think it is better to remove it for now, we can always re-implement it if it turns out to be necessary.

march_hardware/src/EtherCAT/EthercatMaster.cpp
```diff
@@ -96,7 +97,7 @@ void EthercatMaster::ethercatSlaveInitiation(std::vector<Joint>& joints)
     {
       ec_slave[joint.getIMotionCubeSlaveIndex()].PO2SOconfig = setSlaveWatchdogTimer;
     }
-    joint.initialize(this->cycle_time_ms_);
+    reset |= joint.initialize(this->cycle_time_ms_);
```
Roelemans, line 100:
```suggestion
    reset = joint.initialize(this->cycle_time_ms_);
```
Why not this?

march_hardware/src/IMotionCube.cpp
```diff
@@ -84,9 +84,23 @@ void IMotionCube::mapMosiPDOs()
...
+  int checksum_verified = verifySetup();
+
+  if (!checksum_verified)
+  {
+    ROS_WARN("The .sw file for slave %d is not equal to the setup of the drive, downloading is necessary",
+             this->slaveIndex);
+    downloadSetupToDrive();
+    checksum_verified = verifySetup();
+    ROS_INFO(checksum_verified ? "writing of the setup data has succeeded" : "writing of the setup data has failed");
+    return true;  // Resets the etherCAT train and imc, necessary after downloading a "new" setup to the drive
```
Roelemans:
```suggestion
    return true;  // Resets all imcs and restart etherCAT train (necessary after downloading a "new" setup to the drive)
```
Wouldn't it make more sense to return `false` when it fails and `true` when it succeeds?

march_hardware/src/MarchRobot.cpp
Joris-JW: Is there a situation where it would not work the first time but it does the second time? I guess if etherCAT does not start successfully the first time we should not try this again but display the error which has occurred?

march_hardware/src/IMotionCube.cpp
```diff
@@ -364,6 +380,84 @@ uint32_t IMotionCube::computeSWCheckSum(int& start_address, int& end_address)
                                  delimiter);
 }
 
+int IMotionCube::verifySetup()
+{
```
Roelemans, line 392: Maybe put a reference to documentation explaining how this works?

march_hardware/src/IMotionCube.cpp
```diff
@@ -364,6 +380,84 @@ uint32_t IMotionCube::computeSWCheckSum(int& start_address, int& end_address)
...
+    throw error::HardwareException(error::ErrorType::WRITING_INITIAL_SETTINGS_FAILED,
+                                   "Failed checking the checksum on slave: %d", this->slaveIndex);
+  }
+
+  ROS_DEBUG("The .sw checksum is : %d, and the drive checksum is %d", sw_value, imc_value);
+  return sw_value == imc_value;
+}
+
+void IMotionCube::downloadSetupToDrive()
+{
```
Roelemans, line 414: Here too; maybe put a reference to documentation explaining how this works?

Roelemans (changes requested):

march_hardware/include/march_hardware/IMotionCube.h
RPKrijnen: Okay, if you both think this would be neater I'll change these return values.

march_hardware/src/IMotionCube.cpp
RPKrijnen: fair point. I'll implement these changes.

march_hardware/src/MarchRobot.cpp
RPKrijnen: I guess I'll remove this for now

march_hardware/src/IMotionCube.cpp
RPKrijnen: Do you not think the docstring in the .h file is sufficient?

march_hardware/src/EtherCAT/EthercatMaster.cpp
RPKrijnen: This has already been implemented as of writing this reply

march_hardware/src/IMotionCube.cpp
RPKrijnen: It is coupled to a reset argument, so from that perspective it makes sense to read: my reset argument is True and therefore I must reset.

march_hardware/src/IMotionCube.cpp
```diff
@@ -364,6 +388,84 @@ uint32_t IMotionCube::computeSWCheckSum(int& start_address, int& end_address)
                                  delimiter);
 }
 
+bool IMotionCube::verifySetup()
+{
+  int start_address = 0, end_address = 0;
+  uint32_t sw_value = this->computeSWCheckSum(start_address, end_address);
+  // set parameters to compute checksum on the imc
+  int checksum_setup =
+      sdo_bit32_write(slaveIndex, 0x2069, 0, end_address * 65536 + start_address);  // Endaddress leftshifted 4 times
```
Roelemans:
```suggestion
      sdo_bit32_write(slaveIndex, 0x2069, 0, end_address << 16 + start_address);  // Endaddress leftshifted 4 times
```
I believe you can also do it something like this

march_hardware/src/IMotionCube.cpp
```diff
@@ -364,6 +388,84 @@ uint32_t IMotionCube::computeSWCheckSum(int& start_address, int& end_address)
...
+  size_t pos = 0;
+  size_t old_pos = 0;
+  std::string delimiter = "\n";
+  std::string token, next_token;
+  while ((pos = sw_string_.find(delimiter, old_pos)) != std::string::npos)
+  {
+    token = sw_string_.substr(old_pos, pos - old_pos);
+    if (old_pos == 0)
+    {
+      mem_location = std::stoi(token, nullptr, 16) * hex_ls_four;
```
Roelemans:
```suggestion
      mem_location = std::stoi(token, nullptr, 16) << 16;
```
Same here

Roelemans (changes requested):

march_hardware/src/IMotionCube.cpp
Roelemans: Hmm that docstring sufficiently explains what the method does, but not how it does that. You write a lot of registers and values, so I think a link to IMC documentation explaining how this works (i.e. what value you need to write to which register) would make it easier for subsequent generations to understand and make alterations where necessary. Does that make sense?

march_hardware/src/IMotionCube.cpp
Roelemans: Hmm I believe this is usually done the other way around. You can always do 
`reset = !this->imc_->writeInitialSDOs(cycle_time);`
but you may also keep it as it is, if you prefer.

march_hardware/src/IMotionCube.cpp
RPKrijnen: I'll provide a reference to the CoE manual and the appropriate chapter and paragraph wherein the process is described in detail.

RutgerVanBeek (dismissed):

march_hardware/include/march_hardware/IMotionCube.h
```diff
@@ -99,12 +99,24 @@ class IMotionCube : public Slave
...
+   * This method makes use of the computeSWCheckSum(int&, int&) method. The start and end addresses are used in
+   * conjunction with the registers 0x2069 and 0x206A (described in the CoE manual fro Technosoft(2019) in par. 16.2.5
+   * and 16.2.6) to determine the checksum on the drive.
+   */
+  bool verifySetup();
+  /**
+   * Downloads the setup on the .sw file onto the drive using 32bit SDO write functions.
+   * The general process is specified in chapter 16.4 of the CoE manual from Technosoft(2019).
+   */
+  void downloadSetupToDrive();
```
Olavhaasie: I think it would be nice to make these methods `private`, since they should only be called from `writeInitialSettings` by the `IMotionCube` itself. This also prevents people from mistakenly calling the wrong methods. We can always make them `public` later if necessary, the other way around can be a lot harder.

march_hardware/include/march_hardware/Slave.h
```diff
@@ -30,8 +30,9 @@ class Slave
 
   virtual ~Slave() noexcept = default;
 
-  virtual void writeInitialSDOs(int /* cycle_time */)
+  virtual bool writeInitialSDOs(int /* cycle_time */)
```
Olavhaasie, line 33: Perhaps it would be nice to document the expected return value here? When is it `true` and when is it `false`? You can do so in docstring
```cpp
/**
 * Writes initial SDOs.
 *
 * @return True when ..., false when...
 */

File: march_hardware/src/IMotionCube.cpp [hunk 15]
```diff
@@ -364,6 +388,83 @@ uint32_t IMotionCube::computeSWCheckSum(int& start_address, int& end_address)
                                  delimiter);
 }
 
+bool IMotionCube::verifySetup()
+{
+  int start_address = 0, end_address = 0;
+  uint32_t sw_value = this->computeSWCheckSum(start_address, end_address);
+  // set parameters to compute checksum on the imc
+  int checksum_setup =
+      sdo_bit32_write(slaveIndex, 0x2069, 0, (end_address << 16) + start_address);  // Endaddress leftshifted 4 times
```
Olavhaasie: This seems a useless comment, describing exactly what is happening.
```suggestion
      sdo_bit32_write(slaveIndex, 0x2069, 0, (end_address << 16) + start_address);
```

Olavhaasie (changes requested):

march_hardware/include/march_hardware/Slave.h
RPKrijnen: I will write a short docstring about this function

march_hardware/include/march_hardware/IMotionCube.h
RPKrijnen: These are already private methods?

Olavhaasie: Oh lol, the `private` declaration was just outside the diff

Olavhaasie (dismissed):

Olavhaasie (approved):

RutgerVanBeek (approved):
I added the test-label again, because of the many changes

Roelemans (approved):

PR #258, opened 2020-04-24
Velocity from iMC in pdo and controller

RutgerVanBeek:
Closes PM-415 and PM-423

## Description

I am pretty sure I fixed the issues (coding mistake and fixed-point instead of integer). 
This adds the velocity both measured by the incremental and absolute encoder to the PDO and adds them to the `ImcState` message. 

This part was tested successfully. I have added that this velocity is also used by the controller. The controller now uses the velocity measured by the most accurate encoder. 
## Changes

* Added `MotorVelocity` and `ActualVelocity` to the PDO.
* Adjusted Encoders, Joints and iMC to pass the values of the velocities of the incremental and absolute encoders around.
* Changed the `ImcState` message and update this message with the velocities. 
* Changed `readEncoders` function in `Joint.cpp` to use the velocity from the most accurate encoder
*Changed tests to mock the `readVelocity` functions.


Joris-JW (dismissed):
Your code is looking good, eager to know if this works properly. However, u do got to fix your build and clang formatting ofc!

march_hardware/src/encoder/Encoder.cpp
```diff
@@ -21,6 +21,16 @@ int32_t Encoder::getAngleIU(uint8_t byte_offset) const
...
 }
 
+float Encoder::getVelocityIU(uint8_t byte_offset) const
+{
+    if (this->slave_index_ == -1)
+    {
+        ROS_FATAL("Encoder has slaveIndex of -1");
+    }
+    union bit32 return_byte = get_input_bit32(this->slave_index_, byte_offset);
+    return return_byte.i / 65536.0;
```
Olavhaasie: This seems a lot like a magic number and I think it won't be immediately clear what you mean by this when others are reading this. So, you should define a named constant and perhaps add comment how this should be interpreted.

Olavhaasie (changes requested):

RutgerVanBeek:
The tests showed that the reading of the PDO works. We could merge it without using it in the PDO, or we could continue working on this branch for using it in the controller. The last commit did exactly that, but it was not entirely successful.

Joris-JW:
is the 'not entirely successful' the increase in speed of the exoskeleton? maybe put this PR back to draft mode.

RutgerVanBeek:
> is the 'not entirely successful' the increase in speed of the exoskeleton? maybe put this PR back to draft mode.

There are some small mistakes. I already found out why the speed increased that much. Going back to draft is a good idea.

march_hardware/src/encoder/Encoder.cpp
```diff
@@ -21,6 +21,21 @@ int32_t Encoder::getAngleIU(uint8_t byte_offset) const
...
 }
 
+float Encoder::getVelocityIU(uint8_t byte_offset) const
+{
+  if (this->slave_index_ == -1)
+  {
+    ROS_FATAL("Encoder has slaveIndex of -1");
+  }
+  union bit32 return_byte = get_input_bit32(this->slave_index_, byte_offset);
+  return return_byte.i / FIXED_POINT_TO_FLOAT_CONVERSION;
```
Roelemans:
```suggestion
  return return_byte.i << 16;
```
You can use bit shift operator here

march_hardware/src/IMotionCube.cpp
```diff
@@ -69,6 +69,8 @@ void IMotionCube::mapMisoPDOs()
   map_miso.addObject(IMCObjectName::DetailedErrorRegister);
   map_miso.addObject(IMCObjectName::DCLinkVoltage);
   map_miso.addObject(IMCObjectName::MotorPosition);
+  map_miso.addObject(IMCObjectName::MotorVelocity);
+  map_miso.addObject(IMCObjectName::ActualVelocity);
```
Roelemans, line 73: Why not call these IncrementalVelocity and AbosluteVelocity?

march_hardware/src/Joint.cpp
```diff
@@ -133,7 +133,7 @@ double Joint::getPosition() const
 
 double Joint::getVelocity() const
 {
-  return this->velocity_;
+  return this->imc_->getVelocityRadIncremental();
```
Roelemans: I think it is better to return this->velocity_ here and calculate the velocity in Joint::readEncoders. Also use whichever encoder has a higher resolution. That way, holes in encoder data are patched with this velocity instead of the velocity as we calculated it previously (see line 120-126)

EDIT you probably were gonna fix this anyway :P

Roelemans (changes requested):
Whoops just realised you did not implement the fixes yet oops. Well got some comments that may be useful anyhow

march_hardware/src/encoder/Encoder.cpp
RutgerVanBeek: Here you just throw away the 16 first bits. I will initialize the constant with bit shifts.

march_hardware/src/IMotionCube.cpp
RutgerVanBeek: To have consistency with the absolute and incremental encoders. I can adjust the names, but I suggest we also change the ones of the positions.

march_hardware/src/Joint.cpp
RutgerVanBeek: Yes, I just did.

march_hardware/include/march_hardware/encoder/Encoder.h
```diff
@@ -29,6 +29,21 @@ class Encoder
    */
   double getAngleRad(uint8_t byte_offset) const;
 
+  /**
+   * Reads out the velocity of the encoder from the slave and returns the value in Internal Units (IU).
+   * (IU changed in last second)
+   *  @param byte_offset the byte offset in the slave register for the IU position
+   *  @returns The current velocity measured by the encoder in Internal Units (IU)
+   */
+  float getVelocityIU(uint8_t byte_offset) const;
```
Olavhaasie: I think it would be best to change every `float` to `double`. Currently, you have some velocities `float` and some `double`, while there is no real advantage of using `float` (unless you have a 32 bit system). And in the end it is used by the controllers as a `double`.

march_hardware/src/encoder/Encoder.cpp
```diff
@@ -21,6 +21,21 @@ int32_t Encoder::getAngleIU(uint8_t byte_offset) const
...
 }
 
+float Encoder::getVelocityIU(uint8_t byte_offset) const
+{
+  if (this->slave_index_ == -1)
+  {
+    ROS_FATAL("Encoder has slaveIndex of -1");
+  }
+  union bit32 return_byte = get_input_bit32(this->slave_index_, byte_offset);
+  return return_byte.i / FIXED_POINT_TO_FLOAT_CONVERSION * TIME_PER_VELOCITY_SAMPLE;
```
Olavhaasie: You can effortlessly change the constants to `double` here and the result will be a `double` as well.

Olavhaasie (changes requested):
As I said in my comments, take a look at changing all `float` to `double`.

march_hardware/include/march_hardware/encoder/Encoder.h
```diff
@@ -29,6 +29,21 @@ class Encoder
...
+   * (IU changed in last second)
+   *  @param byte_offset the byte offset in the slave register for the IU position
+   *  @returns The current velocity measured by the encoder in Internal Units (IU)
+   */
+  double getVelocityIU(uint8_t byte_offset) const;
+
+  /**
+   * Reads out the velocity of the encoder from the slave and transforms the result to radians.
+   * @param byte_offset the byte offset in the slave register for the IU position
+   * @returns The current position of the encoder in radians
```
Olavhaasie:
```suggestion
   * @returns The current velocity measured by the encoder in radians
```

march_hardware/include/march_hardware/encoder/Encoder.h
```diff
@@ -51,6 +66,12 @@ class Encoder
   static const size_t MIN_RESOLUTION = 1;
   static const size_t MAX_RESOLUTION = 32;
 
+  // constant used for converting a fixed point 16.16 bit number to a double, which is done by dividing by 2^16
+  static constexpr double FIXED_POINT_TO_FLOAT_CONVERSION = float(1 << 16);
```
Olavhaasie:
```suggestion
  static constexpr double FIXED_POINT_TO_FLOAT_CONVERSION = 1 << 16;
```

Olavhaasie (changes requested):

march_hardware/include/march_hardware/encoder/Encoder.h
```diff
@@ -29,6 +29,21 @@ class Encoder
    */
   double getAngleRad(uint8_t byte_offset) const;
 
+  /**
+   * Reads out the velocity of the encoder from the slave and returns the value in Internal Units (IU).
+   * (IU changed in last second)
+   *  @param byte_offset the byte offset in the slave register for the IU position
+   *  @returns The current velocity measured by the encoder in Internal Units (IU)
+   */
```
Roelemans:
```suggestion
   * Reads out the velocity of the encoder from the slave and returns the value in Internal Units per second (IU/s).
   *  @param byte_offset the byte offset in the slave register for the IU velocity
   *  @returns The current velocity measured by the encoder in Internal Units per second (IU/s)
   */
```

march_hardware/include/march_hardware/encoder/Encoder.h
Roelemans:
```suggestion
   * Reads out the velocity of the encoder from the slave and transforms the result to radians / sec.
   * @param byte_offset the byte offset in the slave register for the IU velocity
   * @returns The current position of the encoder in radians / sec
```

march_hardware/include/march_hardware/encoder/Encoder.h
```diff
@@ -51,6 +66,12 @@ class Encoder
   static const size_t MIN_RESOLUTION = 1;
   static const size_t MAX_RESOLUTION = 32;
 
+  // constant used for converting a fixed point 16.16 bit number to a double, which is done by dividing by 2^16
+  static constexpr double FIXED_POINT_TO_FLOAT_CONVERSION = float(1 << 16);
+
+  // iMOTIONCUBE setting (slow loop sample period
```
Roelemans:
```suggestion
  // iMOTIONCUBE setting (slow loop sample period)
```

march_hardware/src/Joint.cpp
```diff
@@ -95,27 +95,21 @@ void Joint::readEncoders(const ros::Duration& elapsed_time)
...
+      this->position_ = this->imc_->getAngleRadAbsolute();
     }
-
-    // Update position with the most accurate velocity
-    this->position_ += best_displacement;
-    this->velocity_ = best_displacement / elapsed_time.toSec();
-    this->incremental_position_ = new_incremental_position;
-    this->absolute_position_ = new_absolute_position;
+    this->incremental_position_ += incremental_position_change;
+    this->absolute_position_ = this->imc_->getAngleRadAbsolute();
```
Roelemans, line 112: You could consider removing absolute_position, getIncrementalPosition and getAbsolutePosition because they are only used in tests

march_hardware/src/encoder/Encoder.cpp
```diff
@@ -21,6 +21,21 @@ int32_t Encoder::getAngleIU(uint8_t byte_offset) const
...
 }
 
+double Encoder::getVelocityIU(uint8_t byte_offset) const
+{
+  if (this->slave_index_ == -1)
+  {
+    ROS_FATAL("Encoder has slaveIndex of -1");
+  }
+  union bit32 return_byte = get_input_bit32(this->slave_index_, byte_offset);
+  return return_byte.i / FIXED_POINT_TO_FLOAT_CONVERSION * TIME_PER_VELOCITY_SAMPLE;
```
Roelemans: Does the IMotionCube return velocity IU/ms? Why not just IU/s :upside_down_face:

Roelemans:
```suggestion
  return return_byte.i * TIME_PER_VELOCITY_SAMPLE / FIXED_POINT_TO_FLOAT_CONVERSION;
```
I somewhat prefer this order to avoid any ambiguity in order of calculations.

Roelemans (changes requested):
The only place elapsed_time is used now is in line 116-119 of Joint.cpp, when joints did not receive a data update. In this case you should just go with the most accurate time since the last measurement, so you should remove the min in line 61 of march_hardware_interface_node.cpp.

march_hardware/src/Joint.cpp
RutgerVanBeek: I think when this whole thing with the encoders is finished it might be nice, to review what is necessary and what not and what code optimizations we can do. But if we do that before we have a satisfying result we might need them later on.

march_hardware/src/encoder/Encoder.cpp
RutgerVanBeek: It returns the number of bits moved in the sample period which defaults to 1 ms.

Olavhaasie: Better yet is to add braces

march_hardware/src/Joint.cpp
Joris-JW: I suspect it wont be long before we transition to the new Ubuntu. This transition requires a lot of refactoring and stuff so we could always do it during this phase.

Joris-JW (dismissed):

march_hardware_interface/src/march_hardware_interface_node.cpp
```diff
@@ -58,7 +58,7 @@ int main(int argc, char** argv)
       march.waitForPdo();
 
       const ros::Time now = ros::Time::now();
-      ros::Duration elapsed_time = std::min(now - last_update_time, minimal_elapsed_time);
+      ros::Duration elapsed_time = now - last_update_time;
```
Roelemans, line 60: Also remove the minimal_elapsed_time variable? (You can just initialize the last_update_time to now)

Roelemans (changes requested):

march_hardware_interface/src/march_hardware_interface_node.cpp
```diff
@@ -48,8 +48,7 @@ int main(int argc, char** argv)
   }
 
   controller_manager::ControllerManager controller_manager(&march, nh);
-  const ros::Duration minimal_elapsed_time = ros::Duration(march.getEthercatCycleTime() / 1000.0);
-  ros::Time last_update_time = ros::Time::now() - minimal_elapsed_time;
+  ros::Time last_update_time = ros::Time::now()
```
Roelemans:
```suggestion
  ros::Time last_update_time = ros::Time::now();
```
;) Also nice commit name

Roelemans (changes requested):
Buce

march_hardware_interface/src/march_hardware_interface_node.cpp
RutgerVanBeek: oh shit

Roelemans (approved):

Olavhaasie (approved):

PR #259, opened 2020-04-27
Fatal when safety limits act from start

RutgerVanBeek:
Closes PM-394

## Description

When the safety limits act on start-up there is probably something going wrong. Either the encoders are read wrong or the joint is in a strange position (probably outside soft limits). We don't want to continue. This PR checks the value of the `joint_effort_command_` after the safety limits have been enforced. A fatal is thrown when a non-zero effort is found. 

I still think we should also check whether a joint is inside the soft-limits during start-up, because that would prevent many situation in which this check would work. 
## Changes

Added a check for non-zero efforts the first time we want to actuate


march_imu_manager/src/wireless_master.cpp
```diff
@@ -87,8 +87,8 @@ int WirelessMaster::configure(const int update_rate, const int channel)
   }
   ROS_DEBUG_STREAM("Supported update rates: " << update_rates.str());
 
-  const int new_update_rate = findClosestUpdateRate(supported_update_rates, update_rate);
-
```
Olavhaasie: I think this an oopsie

Olavhaasie (changes requested):
You say this happens when the joints are outside of its limits? So wouldn't it be better to just make sure the joints are not outside soft limits during initialization? Furthermore, I have seen this happen before, where the joints began moving, only that happened after a few cycles and not at the first.

RutgerVanBeek:
> You say this happens when the joints are outside of its limits? So wouldn't it be better to just make sure the joints are not outside soft limits during initialization? Furthermore, I have seen this happen before, where the joints began moving, only that happened after a few cycles and not at the first.

We also had this, when we tried something with reading the encoders and made a mistake. I think we should check for the soft limits, but we don't know what we implemented in the future. In any case I still think you should not continue startup, when the safety limits start working directly. 

Are you sure they started after a few cycles? Couldn't there be a small delay in applying a large effort and the motor actually having a large velocity. In the bag files I have looked there was a large effort direct from start-up. How would you change the implementation, would you check until a homing gait is selected?

Olavhaasie:
> We also had this, when we tried something with reading the encoders and made a mistake.

This should be able to be prevented by checking the limits right? Right now the implementation just checks for output of the safety limits and only reports `ROS_FATAL`, so this won't help prevent your mistakes and will continue moving.

> I think we should check for the soft limits, but we don't know what we implemented in the future. 

What do you mean by this? Why can we not implement it right now? Currently, it only checks for hard limits but can easily be extended to soft limits.

> In any case I still think you should not continue startup, when the safety limits start working directly.

This is not necessarily true in every case. If joints are allowed to be outside soft limits on startup, the joints can be actuated to just outside the soft limits. Seems to me an easy way of automatically adjusting the joints. We might not want to move at startup for the pilot. Therefore, it is better to check for soft limits.

> Are you sure they started after a few cycles? Couldn't there be a small delay in applying a large effort and the motor actually having a large velocity. In the bag files I have looked there was a large effort direct from start-up.

Could you provide the bag files here?

RutgerVanBeek:
> What do you mean by this? Why can we not implement it right now? Currently, it only checks for hard limits but can easily be extended to soft limits.

We can implement this, but we don't know what kind of issues we might have in the future. They shouldn't be there, but as a safety measure we could also have this. 

> This is not necessarily true in every case. If joints are allowed to be outside soft limits on startup, the joints can be actuated to just outside the soft limits. Seems to me an easy way of automatically adjusting the joints. We might not want to move at startup for the pilot. Therefore, it is better to check for soft limits.

In one of the bag files (the one with the right ankle wiggle) there is no outside soft limits reported. (At least I couldn't find it on ROS out). I did not expect that, but that means we have more issues 

Yes, you are right, but letting the safety limits acts is not a good way. It might be a good idea to turn off the safety limits before you have homed. 


[log_2020-04-22-14-05-24.zip](https://github.com/project-march/hardware-interface/files/4540204/log_2020-04-22-14-05-24.zip)
[log_2020-04-22-14-11-20.zip](https://github.com/project-march/hardware-interface/files/4540208/log_2020-04-22-14-11-20.zip)

RutgerVanBeek:
I just realized that although we do not check the soft limits on start-up it is always done before we write. In the `march_hardware_interface_node.cpp` the order is to read, to validate (outside limits check), to update the controller and then to write. So the soft limits are checked before we write. Another thing that is strange is that we write although in ros bag it looks like the controller is not finished initializing.

RutgerVanBeek:
It is changed to check until the controller starts actuating (probably a homing gait).

Joris-JW:
> In one of the bag files (the one with the right ankle wiggle) there is no outside soft limits reported. (At least I couldn't find it on ROS out). I did not expect that, but that means we have more issues

But if this is the case and no soft limits are surpassed, the enforceLimits has no influence right? If so, I agree that we have more issues but are they all related to the enforcement of the soft limit functionalities?

Joris-JW (dismissed):
I agree with Olav on checking the soft limits as an additional check. However, I do think this safety feature is a nice touch, it should prevent unwanted movements during startup originating from who knows where. Start up wiggles should be 100& terminated if this works properly.

Joris-JW (approved):

Roelemans (approved):

Olavhaasie (approved):

PR #261, opened 2020-05-05
set joint_state_publisher frequency to 250 Hz

RutgerVanBeek:
## Description

The ethercat loop is running at 250 Hz. The positions, velocities and efforts are thus updatet at 250 Hz, yet we published them only at 50 Hz. Rogier and I want to look at the acceleration and effort and then we also need the acceleration with the previous effort.
## Changes

Changed the publish frequency of the `joint_state_publisher` from 50 to 250 Hz.


Olavhaasie (commented):
> Rogier and I want to look at the acceleration and effort and then we also need the acceleration with the previous effort.

What do you mean by this?

Joris-JW:
Do we know why it was set to 50 Hz in the first place? And maybe nice to also change these for the test joint controllers (Otherwise I suspect we will forget to do this later on).

RutgerVanBeek:
> > Rogier and I want to look at the acceleration and effort and then we also need the acceleration with the previous effort.
> 
> What do you mean by this?

That we want to compare the effort the iMC is doing at time k, with the acceleration between time k and k+1. Thus we would need the value of the effort at time k and the velocities at time k and k+1. Currently we only publish one in five values, so we would miss have to compare the effort at time k-5 or at time k with the acceleration at time k. I think this would be less representative.

RutgerVanBeek:
> Do we know why it was set to 50 Hz in the first place? And maybe nice to also change these for the test joint controllers (Otherwise I suspect we will forget to do this later on).

I think almost everything was set to 50 Hz, most robots run on 50 Hz and the `robot_state_publisher` also published at 50 Hz.

Roelemans (commented):
Is it possible to make this directly dependent on the ethercat cycle time?

RutgerVanBeek:
> Is it possible to make this directly dependent on the ethercat cycle time?

I think that is indeed best, but I didn't see a way to do this.

Joris-JW:
> I think that is indeed best, but I did see a way to do this.

Should we wait for this or do you suggest to create a separate issue? Maybe for your tests it's nice to have this PR merged but as Roel suggested the cycle time dependency is better!

RutgerVanBeek:
> > I think that is indeed best, but I did see a way to do this.
> 
> Should we wait for this or do you suggest to create a separate issue? Maybe for your tests it's nice to have this PR merged but as Roel suggested the cycle time dependency is better!

Okay that was a very misleading type, it was supposed to say didn't instead of did. I don't know how you would make it dependent on the ethercat cycle time.

Roelemans (approved):

Roelemans (approved):

Olavhaasie (approved):

PR #262, opened 2020-05-05
Feature: ethercat interface

Olavhaasie:
Closes PM-325

## Description
This PR introduces two new interfaces for reading and writing PDOs and SDOs, named `PdoInterface` and `SdoInterface`. These two interfaces provide functions for reading and writing SDO and PDO types. They are implemented with SOEM global functions in `PdoInterfaceImpl` and `SdoInterfaceImpl`. Since these interfaces provide an interface to all slaves, there also exists a more specific implentation for reading and writing to a single slave, named `PdoSlaveInterface` and `SdoSlaveInterface`. These last two interfaces are passed around where classes need the function to read or write objects. This makes it completely independent from the type of implementation used and hopefully easier to use for developers.

This change also required refactoring `Slave`. `Slave` now inherits from `PdoSlaveInterface` so that the read and write methods can be used by derived classes. Furthermore, the `initSdo` function provides the interface for reading and writing SDOs, so that these objects are restricted to only be written during initialization. This change also required for every slave to have a valid slave index, so I changed the type to `uint16_t`, which disallows negative slave indices. The new interfaces also resulted in slaves not having to know their slave index when calling the read and write methods. This counts for IMC, encoders, pdb, ges and voltage nets. Also the `SdoInterface` handles reading and writing types of every type. The `PdoInterface` still uses the unions.

Small implementation detail of these interfaces means that all interfaces all stored as `std::shared_ptr`, since only one implementation is needed, it is shared among all slaves. This shared implementation instance can be created with `[SP]doInterfaceImpl::create()`. This function is used in the hardware builder.

Finally, this implementation allows for better control during testing using mocks for the interfaces. I have not yet added many tests, but I have added some examples, see `TestEncoder`. This feature required a lot of changes, but I think it is worth it considering the end result.

## Changes
* Add new interfaces: `PdoInterface`, `PdoSlaveInterface`, `SdoInterface`, `SdoSlaveInterface`
* Add SOEM based implementations of of interfaces: `PdoInterfaceImpl` and `SdoInterfaceImpl`
* Remove default constructor on `Slave` and subsequently on `PowerDistributionBoard`
* Rename `writingInitialSettings` to `initSdo`
* Implement `reset` function on `Slave`
* Make PDB optional in `MarchRobot`
* Fix PDB usage in `MarchHardwareInterface`


march_hardware/include/march_hardware/EtherCAT/slave.h
```diff
@@ -0,0 +1,67 @@
+// Copyright 2019 Project March.
```
RutgerVanBeek, line 1: Not specific to this PR, but I notice it now. Should we update the year?

march_hardware/src/PDOmap.cpp
```diff
@@ -96,38 +98,39 @@ std::unordered_map<IMCObjectName, uint8_t> PDOmap::configurePDO(int slave_index,
...
   // Make sure the last PDO is activated
-  sdo_bit8_write(slave_index, current_register, 0, counter - 1);
+  sdo.write<uint8_t>(current_register, 0, counter - 1);
 
   // Deactivated the sync manager and configure with the new PDO
-  sdo_bit8_write(slave_index, base_sync_manager, 0, 0);
-  int currentPDONr = (current_register - base_register) + 1;
-  sdo_bit16_write(slave_index, base_sync_manager, currentPDONr, current_register);
+  sdo.write<uint8_t>(base_sync_manager, 0, 0);
+  int current_pdo = (current_register - base_register) + 1;
```
RutgerVanBeek:
```suggestion
  int current_pdo_number = (current_register - base_register) + 1;
```

RutgerVanBeek (commented):
Almost no comments on the coding. I don't understand yet why the pdo and sdo interfaces are classes. The implementations only really store a slave index. Can't we move the functionality from the implementations to the Slave object. Intuitively the pdo and sdo are not really objects more like a way of communication and they store almost no information. This would remove a level of abstraction. Let me know if I miss a reason to have the interface this way.

march_hardware/include/march_hardware/EtherCAT/slave.h
Olavhaasie: It is not obligatory to have these in the first place. Also it does not seem to be [necessary](https://stackoverflow.com/a/2391555) to update them

RutgerVanBeek: I know, but I we have them and make changes, shouldn't we update them?

Or is it useless to change these kind of things until we have license?

Olavhaasie:
There is two big reasons why I did it this way.

1. We can mock the interfaces
2. Hide implementation and other details

If the functions would be directly implemented on the `Slave` class then there is no way to mock these functions when testing, for example, `IMotionCube`. The interfaces make this a lot easier, in exchange for more abstractions. Furthermore, other classes that use this interface to read and write objects would have to use a `Slave` instance, while only needing to write or read objects. Take for example `Encoder`, this class only needs a reference to an interface that reads and writes objects to a specific slave, nothing else. The interface makes it easy and simple to use and only provides the necessary functions.

RutgerVanBeek (dismissed):

march_hardware/include/march_hardware/EtherCAT/slave.h
Roelemans: According to his link it is better to not update it since it would mean we no longer claim copyright for anything before the new date

march_hardware/include/march_hardware/encoder/AbsoluteEncoder.h
```diff
@@ -36,16 +36,14 @@ class AbsoluteEncoder : public Encoder
...
            lhs.lower_soft_limit_iu_ == rhs.lower_soft_limit_iu_ && lhs.upper_limit_iu_ == rhs.upper_limit_iu_ &&
            lhs.lower_limit_iu_ == rhs.lower_limit_iu_ && lhs.zero_position_iu_ == rhs.zero_position_iu_;
   }
   /** @brief Override stream operator for clean printing */
   friend std::ostream& operator<<(std::ostream& os, const AbsoluteEncoder& encoder)
   {
-    return os << "slaveIndex: " << encoder.getSlaveIndex() << ", "
-              << "totalPositions: " << encoder.getTotalPositions() << ", "
+    return os << "totalPositions: " << encoder.getTotalPositions() << ", "
               << "upperHardLimit: " << encoder.upper_limit_iu_ << ", "
```
Roelemans, line 47: Why did you remove the slaveindex here?

march_hardware/include/march_hardware/Slave.h
```diff
@@ -1,45 +0,0 @@
-// Copyright 2019 Project March.
```
Roelemans, line 1: You emptied this file but failed to delete it ;)

Roelemans (changes requested):
Looked at this for 1.5 hours and still only got some very minor comments :O this is some need code (or I just got delirious from staring at it for too long). 

Very nice changes :) This more object-oriented implementation makes more sense to me

march_hardware/include/march_hardware/encoder/AbsoluteEncoder.h
Olavhaasie: Because the encoders no longer need a slave index in order to read and write PDOs. This now done in `Encoder` class where it accepts a `PdoSlaveInterface`.

march_hardware/include/march_hardware/Slave.h
Olavhaasie: Ah must have happened during the merge with develop :crying_cat_face:

RutgerVanBeek (approved):

Roelemans (approved):

PR #264, opened 2020-05-11
Feature/pm 436 add subindex to pdo mapping

Joris-JW:
Closes PM-439

## Description
Added the sub index to the construction of the combined address which is used when defining the PDO's for every slave. I have chosen to add the sub index to every IMCobject instead of defining a default value. This way it would be most intuitive, The added tests verify that the combined address is constructed properly 

(_for more information about the construction of the PDO combined address look up page 39 in the IMC CoE manual_)

## Changes
* Added the subindex to the constrcutor of the IMCObject
* Added the subindex to the combinedAddress of an IMCObject
* Added tests to verify the combinedAddress

### DIFF
--- a/march_hardware/include/march_hardware/PDOmap.h
+++ b/march_hardware/include/march_hardware/PDOmap.h
@@ -15,14 +15,17 @@ namespace march
 /** Store IMC data as a struct to prevent data overlap.*/
 struct IMCObject
 {
-  uint16_t address;           // in IMC memory (see IMC manual)
-  uint16_t length;            // bits (see IMC manual)
+  uint16_t address;            // in IMC memory (see IMC manual)
+  uint8_t sub_index;          // sub index corresponding to PDO register (see IMC manual)
+  uint8_t length;             // bits (see IMC manual)
   uint32_t combined_address;  // combine the address(hex), sub-index(hex) and length(hex)
 
-  explicit IMCObject(uint16_t _address, uint16_t _length) : address(_address), length(_length)
+  explicit IMCObject(uint16_t _address, uint8_t _sub_index, uint8_t _length)
+    : address(_address), sub_index(_sub_index), length(_length)
   {
-    uint32_t MSword = ((address & 0xFFFF) << 16);  // Shift 16 bits left for most significant word
-    uint32_t LSword = (length & 0xFFFF);
+
+    uint32_t MSword = ((address & 0xFFFF) << 16);
+    uint32_t LSword = ((sub_index & 0xFFFF) << 8) | (length & 0xFFFF);
 
     combined_address = (MSword | LSword);
   }
--- a/march_hardware/include/march_hardware/PDOmap.h
+++ b/march_hardware/include/march_hardware/PDOmap.h
@@ -15,7 +15,7 @@ namespace march
 /** Store IMC data as a struct to prevent data overlap.*/
 struct IMCObject
 {
-  uint16_t address;            // in IMC memory (see IMC manual)
+  uint16_t address;           // in IMC memory (see IMC manual)
   uint8_t sub_index;          // sub index corresponding to PDO register (see IMC manual)
   uint8_t length;             // bits (see IMC manual)
   uint32_t combined_address;  // combine the address(hex), sub-index(hex) and length(hex)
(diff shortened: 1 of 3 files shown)


march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -16,13 +16,15 @@ namespace march
...
   uint32_t combined_address;  // combine the address(hex), sub-index(hex) and length(hex)
 
-  explicit IMCObject(uint16_t _address, uint16_t _length) : address(_address), length(_length)
+  explicit IMCObject(uint16_t _address, uint8_t _sub_index, uint8_t _length)
+    : address(_address), sub_index(_sub_index), length(_length)
   {
-    uint32_t MSword = ((address & 0xFFFF) << 16);  // Shift 16 bits left for most significant word
-    uint32_t LSword = (length & 0xFFFF);
+    uint32_t MSword = ((address & 0xFFFF) << 16);
+    uint32_t LSword = ((sub_index & 0xFFFF) << 8) | (length & 0xFFFF);
```
Roelemans:
```suggestion
    uint32_t MSword = ((address) << 16);
    uint32_t LSword = ((sub_index) << 8) | (length);
```
Isn't this equivalent? (though since it was in there it probably had a use)

Roelemans (dismissed):
One small question. Good job :) (was simpler than I expected haha)

march_hardware/include/march_hardware/PDOmap.h
Joris-JW: silly me

Roelemans (dismissed):

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -16,13 +16,15 @@ namespace march
...
+  uint8_t length;             // bits (see IMC manual)
   uint32_t combined_address;  // combine the address(hex), sub-index(hex) and length(hex)
 
-  explicit IMCObject(uint16_t _address, uint16_t _length) : address(_address), length(_length)
+  explicit IMCObject(uint16_t _address, uint8_t _sub_index, uint8_t _length)
+    : address(_address), sub_index(_sub_index), length(_length)
   {
-    uint32_t MSword = ((address & 0xFFFF) << 16);  // Shift 16 bits left for most significant word
-    uint32_t LSword = (length & 0xFFFF);
+    uint32_t MSword = ((address & 0xFFFF) << 16);
```
Olavhaasie: The bitwise and is actually not necessary, since it is already a 16 bit number.
```suggestion
    uint32_t MSword = (address << 16);
```

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -16,13 +16,15 @@ namespace march
...
+    uint32_t LSword = ((sub_index & 0xFFFF) << 8) | length;
```
Olavhaasie: The value `0xFFFF` used in the bitwise and operator is a 16 bit value, which is bigger than the 8 bit value used, also it is not necessary. I think this makes it easier to read.
```suggestion
    uint32_t LSword = (sub_index << 8) | length;
```

march_hardware/include/march_hardware/PDOmap.h
```diff
@@ -16,13 +16,15 @@ namespace march
...
 {
   uint16_t address;           // in IMC memory (see IMC manual)
-  uint16_t length;            // bits (see IMC manual)
+  uint8_t sub_index;          // sub index corresponding to PDO register (see IMC manual)
+  uint8_t length;             // bits (see IMC manual)
   uint32_t combined_address;  // combine the address(hex), sub-index(hex) and length(hex)
 
-  explicit IMCObject(uint16_t _address, uint16_t _length) : address(_address), length(_length)
+  explicit IMCObject(uint16_t _address, uint8_t _sub_index, uint8_t _length)
+    : address(_address), sub_index(_sub_index), length(_length)
```
Olavhaasie: Now that we are changing this anyway, it would be nice to remove the [`explicit`](https://en.cppreference.com/w/cpp/language/explicit), since that is only meaningful on constructors with one paramater.
```suggestion
  IMCObject(uint16_t _address, uint8_t _sub_index, uint8_t _length)
    : address(_address), sub_index(_sub_index), length(_length)
```

Olavhaasie (changes requested):

march_hardware/include/march_hardware/PDOmap.h
Joris-JW: I guess that would indeed be better

Olavhaasie (approved):

Roelemans (approved):

PR #265, opened 2020-05-12
sort joint names alphabetically in yaml

RutgerVanBeek:
## Description

Lately the sorting of the joints was only done before uploading them to the parameter server (#252). I thought this had no impact, but the joint names are in a different order in the `ImcState.msg` and `AfterLimitJointCommand`. This is a bit unfortunate for the RQT multiplot configurations and the conversion to csv. The easiest was to make sure the joints are used alphabetically again. I tried implementing the sort in the hardware builder, but the happens in place and because of the `unique_ptr` I didn't get it to work. Sorting in the yaml was easier. 
## Changes

* Sort the joint names in the yaml alphabetically

### DIFF
--- a/march_hardware_builder/robots/march4.yaml
+++ b/march_hardware_builder/robots/march4.yaml
@@ -2,17 +2,17 @@ march4:
   ifName: enp2s0
   ecatCycleTime: 4
   joints:
-    - right_hip_aa:
+    - left_ankle:
         actuationMode: torque
         allowActuation: true
-        netNumber: 1
-        temperatureges: {slaveIndex: 7, byteOffset: 4}
+        netNumber: 8
+        temperatureges: {slaveIndex: 1, byteOffset: 0}
         imotioncube:
-          slaveIndex: 8
+          slaveIndex: 2
           absoluteEncoder:
             resolution: 12
-            minPositionIU: 3713
-            maxPositionIU: 4091
+            minPositionIU: 3641
+            maxPositionIU: 4082
           incrementalEncoder:
             resolution: 13
             transmission: 200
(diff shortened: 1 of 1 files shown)


Joris-JW (dismissed):
nice.

march_hardware_builder/robots/march4.yaml
```diff
@@ -107,17 +77,47 @@ march4:
...
+        temperatureges: {slaveIndex: 9, byteOffset: 4}
+        imotioncube:
+          slaveIndex: 11
+          absoluteEncoder:
+            resolution: 17
+            minPositionIU: 41547
+            maxPositionIU: 85287
+          incrementalEncoder:
+            resolution: 12
+            transmission: 101
```
Ishadijcks: Having no new line can break the parsing of YAML files

march_hardware_builder/robots/march4.yaml
```diff
@@ -2,17 +2,17 @@ march4:
   ifName: enp2s0
   ecatCycleTime: 4
   joints:
```
Ishadijcks, line 5: Add a comment here that new joints have to be sorted alphabetically

Ishadijcks (changes requested):
Requesting changes as this could introduce some hard to find bugs. See comments

Must this also be added to other robots (`march3.yaml`) for compatibility?

RutgerVanBeek:
> Requesting changes as this could introduce some hard to find bugs. See comments
> 
> Must this also be added to other robots (`march3.yaml`) for compatibility?

No, because we have no rqt multiplot configurations for the march3. It is not necessarily that it should be alphabetically, but that is what some (non-essential) things assume, liike the rqt muliplot configurations.

Olavhaasie (approved):

Joris-JW (approved):

PR #266, opened 2020-05-12
Feature/pm 422 add motor voltage

Ciliaclaij:
Closes PM-422

## Description
Add the reading of the motor voltage for linear joint control

## Changes
e.g.
* Added motor voltage reading
* Added message to publish motor voltage
* Merged PM-436 into this branch to use subindex PDO reading


march_hardware/src/IMotionCube.cpp
```diff
@@ -296,6 +301,11 @@ float IMotionCube::getIMCVoltage()
          static_cast<float>(imc_voltage_iu);  // Conversion to Volt, see Technosoft CoE programming manual
 }
 
+float IMotionCube::getMotorVoltage()
+{
+  return this->read16(this->miso_byte_offsets_.at(IMCObjectName::DCLinkVoltage)).ui;
```
RutgerVanBeek: DCLinkVoltage is the other voltage or am I wrong? At least this is the same object as used in the `getIMCVoltage()`

RutgerVanBeek (changes requested):
I think you know this better what the `DCLinkVoltage` represents. Currently the `getIMCVoltage()` and `getMotorVoltage()` point both to the `DCLinkVoltage`. You should add an object to the PDO and use that object for one of the two.

march_hardware/src/IMotionCube.cpp
Ciliaclaij: That was a mistake with copying the lines, after fixing all the merge conflicts :(

RutgerVanBeek:
There was an error when writing the sdo's on startup, see Slack for more information. (https://projectmarch.slack.com/archives/CJYPA42UC/p1589362128001700)

Ishadijcks:
Maybe provide a link to the relevant Slack conversation?

RutgerVanBeek (dismissed):

Joris-JW (commented):
I would wait with the merge of this branch until PM 436 is merged to develop, chronological speaking that would be most logical.

RutgerVanBeek:
Oh I remember something now, @Ciliaclaij where you planning on converting the values to more understandable units or is that for another issue?

Ciliaclaij:
@RutgerVanBeek I realised this, when I work again on monday I will implement this. So it's better to wait for that (and PM 436) to merge this

march_hardware/src/IMotionCube.cpp
```diff
@@ -296,6 +302,11 @@ float IMotionCube::getIMCVoltage()
          static_cast<float>(imc_voltage_iu);  // Conversion to Volt, see Technosoft CoE programming manual
 }
 
+float IMotionCube::getMotorVoltage()
+{
+  return this->read16(this->miso_byte_offsets_.at(IMCObjectName::MotorVoltage)).ui;
```
Olavhaasie, line 307: Are you aware that the `.ui` at the end means that the value is read as an unsigned 16 bit integer? I ask this because the return type of the function is `float`. To get the `float` value you would have to do `.f`.

RutgerVanBeek: This is probably the conversion issue I mention above.

RutgerVanBeek: You read only 16bit so you cannot make a float out of it. It is a bit strange that they say you have to read 16 bits, but that it is a float. That seems a mistake on their side.

march_hardware/src/IMotionCube.cpp
```diff
@@ -304,7 +304,7 @@ float IMotionCube::getIMCVoltage()
 
 float IMotionCube::getMotorVoltage()
 {
-  return this->read16(this->miso_byte_offsets_.at(IMCObjectName::MotorVoltage)).ui;
+  return this->read16(this->miso_byte_offsets_.at(IMCObjectName::MotorVoltage)).f;
```
Olavhaasie: As Rutger already mentioned, this is not possible since float can only be a 32 bit value. You would have to do some manual conversions. You should convert this back for now and we'll see later about the conversion.
```suggestion
  return this->read16(this->miso_byte_offsets_.at(IMCObjectName::MotorVoltage)).ui;
```

Olavhaasie (approved):

RutgerVanBeek (approved):

PR #267, opened 2020-05-18
Use sw file from ems repository

Roelemans:
Dependent on https://github.com/project-march/ems-projects/pull/1
Closes PM-429

## Description

Get sw files from the ems-projects repo instead of the march_hardware_interface.


RutgerVanBeek (dismissed):

Olavhaasie (changes requested):
you should also remove the `config` directory from `install` command
https://github.com/project-march/hardware-interface/blob/d7f6e83b9f8038b8a5253053e61bbff5f75611b6/march_hardware_builder/CMakeLists.txt#L40

RutgerVanBeek (approved):
Tested works well, don't forget to push the naming mistake in the ems projects repo

Olavhaasie (approved):
Good :+1: It would be nice to change the hardcoded package and path to an option the YAML sometime in the future, to avoid having to install the package when writing tests for example.

PR #269, opened 2020-05-28
Feature/pm 301 voltage to velocity

Ciliaclaij:
Closes PM-301

## Description

Added function to get the motor velocity out of the motor voltage and motor current.

### DIFF
--- a/march_hardware/src/joint.cpp
+++ b/march_hardware/src/joint.cpp
@@ -131,6 +131,13 @@ double Joint::getVelocity() const
   return this->velocity_;
 }
 
+float Joint::getVoltageVelocity() const
+{
+  float R = 0.05;
+  float Ke = 355 * M_PI / 30;
+  return (this->imc_->getMotorVoltage() + this->imc_->getMotorCurrent() * R) / Ke;
+}
+
 double Joint::getIncrementalPosition() const
 {
   return this->incremental_position_;
--- a/march_hardware/src/joint.cpp
+++ b/march_hardware/src/joint.cpp
@@ -133,11 +133,12 @@ double Joint::getVelocity() const
 
 double Joint::getVoltageVelocity() const
 {
-  double R_motor = 0.05;
-  double Kv_motor = 355;
-  double rpmToRad = M_PI / 30;
-  double Ke = Kv_motor * rpmToRad;
-  return (this->imc_->getMotorVoltage() + this->imc_->getMotorCurrent() * R_motor) / Ke;
+  // https://en.wikipedia.org/wiki/Motor_constants#Motor_velocity_constant,_back_EMF_constant
+  const double resistance = 0.05;
+  const double velocity_constant = 355;
+  const double rpm_to_rad = M_PI / 30;
+  const double electric_constant = velocity_constant * rpm_to_rad;
+  return (this->imc_->getMotorVoltage() + this->imc_->getMotorCurrent() * resistance) / electric_constant;
 }
 
 double Joint::getIncrementalPosition() const
--- a/march_hardware/src/joint.cpp
+++ b/march_hardware/src/joint.cpp
@@ -133,7 +133,7 @@ double Joint::getVelocity() const
 
 double Joint::getVoltageVelocity() const
 {
-  // For the underlying calculations, see: 
+  // For the underlying calculations, see:
   // https://en.wikipedia.org/wiki/Motor_constants#Motor_velocity_constant,_back_EMF_constant
   // https://www.motioncontroltips.com/faq-whats-relationship-voltage-dc-motor-output-speed/
   const double resistance = 0.05;
--- a/march_hardware/src/joint.cpp
+++ b/march_hardware/src/joint.cpp
@@ -133,6 +133,7 @@ double Joint::getVelocity() const
 
 double Joint::getVoltageVelocity() const
 {
+  // For the underlying calculations, see: 
   // https://en.wikipedia.org/wiki/Motor_constants#Motor_velocity_constant,_back_EMF_constant
   // https://www.motioncontroltips.com/faq-whats-relationship-voltage-dc-motor-output-speed/
   const double resistance = 0.05;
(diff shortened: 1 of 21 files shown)


march_hardware/src/joint.cpp
```diff
@@ -131,6 +131,15 @@ double Joint::getVelocity() const
...
 }
 
+float Joint::getVoltageVelocity() const
+{
+  float R_motor = 0.05;
+  float Kv_motor = 355;
+  float rpmToRad = M_PI / 30;
+  float Ke = Kv_motor * rpmToRad;
+  return (this->imc_->getMotorVoltage() + this->imc_->getMotorCurrent() * R_motor) / Ke;
+}
```
Olavhaasie: Usually it is best to use `double` instead of `float`. This gives higher precision for (almost) no additional cost. Also because all inputs for controllers are `double`. It doesn't matter if `getMotorVoltage` and `getMotorCurrent` return `float`, since those will be automatically converted to `double`.

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -195,6 +195,7 @@ void MarchHardwareInterface::read(const ros::Time& /* time */, const ros::Durati
     joint.readEncoders(elapsed_time);
     joint_position_[i] = joint.getPosition();
     joint_velocity_[i] = joint.getVelocity();
+    joint_voltage_velocity_[i] = joint.getVoltageVelocity();
```
Olavhaasie: Before assigning something to a vector like this it has to be initialized. You can do this in the `reserveMemory` method. Is this member variable really necessary? Since it does not serve as input to a controller and will probably only be used as input to a Kalman filter right?

Olavhaasie (changes requested):

march_hardware/src/joint.cpp
```diff
@@ -131,6 +131,15 @@ double Joint::getVelocity() const
   return this->velocity_;
 }
 
+double Joint::getVoltageVelocity() const
+{
+  double R_motor = 0.05;
+  double Kv_motor = 355;
+  double rpmToRad = M_PI / 30;
+  double Ke = Kv_motor * rpmToRad;
+  return (this->imc_->getMotorVoltage() + this->imc_->getMotorCurrent() * R_motor) / Ke;
```
Roelemans:
```suggestion
  double resistance_motor = 0.05;
  double Kv_motor = 355;
  double rpm_to_rad = M_PI / 30;
  double Ke = Kv_motor * rpm_to_rad;
  return (this->imc_->getMotorVoltage() + this->imc_->getMotorCurrent() * resistance_motor) / Ke;
```
Non constant variables are usually named with non-capitals and underscores ;) what are Ke and Kv? maybe write out the abbreviations?

Roelemans: Maybe add a reference or link here to where this calculation comes from?

Roelemans (changes requested):

march_hardware/src/joint.cpp
Ciliaclaij: @Roelemans https://en.wikipedia.org/wiki/Motor_constants#Motor_velocity_constant,_back_EMF_constant This is the information about the Kv and Ke value. The reason I put both of them in there is because we got the Kv_motor directly from the spec sheet of the motor and I need Ke in the derivation in line 140. We can change it to backemf_motor_constant, but I am not sure if this is the better way. Maybe stay with the Kv and Ke and add the wikipedia link?

Ciliaclaij: I have this link that explains the calculation: https://www.motioncontroltips.com/faq-whats-relationship-voltage-dc-motor-output-speed/. Should I add this one or a confluence page where I explain it?

Ciliaclaij: About 'non constant variables', I wonder whether the first four variables are constants. Because they are pretty constant, right?

Olavhaasie: Yes they are very constant. According to [ROS C++ style guide](https://wiki.ros.org/CppStyleGuide#Constants) they should be CAPITAL_SNAKE_CASE. However, since we only do this for global constants I think they should be normal snake_case

```suggestion
  // https://en.wikipedia.org/wiki/Motor_constants#Motor_velocity_constant,_back_EMF_constant
  const double resistance = 0.05;
  const double velocity_constant = 355;
  const double rpm_to_rad = M_PI / 30;
  const double electric_constant = velocity_constant * rpm_to_rad;
  return (this->imc_->getMotorVoltage() + this->imc_->getMotorCurrent() * resistance) / electric_constant;
```
Personally I would do it like this, but you can also choose to keep the `kv` and `ke` names, since those better reflect the theory. Most important is **keep it consistent**. You know variable naming is one of the hardest things in programming :wink:

Normally hardcoding constants is not approved, since these values depend on the motor you are using. Perhaps we should add motor parameters to the robot yaml, but that is something for another time.

Ciliaclaij: @Roelemans ?

Roelemans: That link is fine :+1:

Roelemans: That link is fine :+1:

march_hardware/src/joint.cpp
```diff
@@ -131,6 +131,17 @@ double Joint::getVelocity() const
   return this->velocity_;
 }
 
+double Joint::getVoltageVelocity() const
+{
+  // https://en.wikipedia.org/wiki/Motor_constants#Motor_velocity_constant,_back_EMF_constant
```
Roelemans, line 137: One last thing :) 
```suggestion
  // For the underlying calculations, see: 
  // https://en.wikipedia.org/wiki/Motor_constants#Motor_velocity_constant,_back_EMF_constant
```

Roelemans (changes requested):

Roelemans (dismissed):
Good job :D :tada:

Olavhaasie (dismissed):
Lookin very fine :ok_man:

march_hardware/src/joint.cpp
```diff
@@ -131,6 +131,18 @@ double Joint::getVelocity() const
   return this->velocity_;
 }
 
+double Joint::getVoltageVelocity() const
+{
+  // For the underlying calculations, see:
```
Olavhaasie: sorry :sob: 
```suggestion
  // For the underlying calculations, see:
```

Olavhaasie (approved):

Roelemans (approved):

PR #273, opened 2020-06-16
Fix/voltage velocity reading calculation

Ciliaclaij:
## Description

Changed motor voltage reading to signed integer instead of unsigned integer. Also fixed faulty calculation of the velocity from the voltage. 

## Changes

* Changed motor voltage reading to signed integer instead of unsigned integer
* Fixed calculation of the velocity from voltage

### DIFF
--- a/march_hardware/include/march_hardware/imotioncube/imotioncube.h
+++ b/march_hardware/include/march_hardware/imotioncube/imotioncube.h
@@ -60,9 +60,9 @@ public:
 
   ActuationMode getActuationMode() const;
 
-  virtual float getMotorCurrent();
+  virtual double getMotorCurrent();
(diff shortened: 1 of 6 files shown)


RutgerVanBeek (dismissed):
I would test during the software tests, whether using signed instead of unsigned fixes it perfectly or whether there is still a conversion missing.

Olavhaasie (commented):
The changes look good. How are the results?

Ciliaclaij:
I don't have the results yet :( Rutger has not been able yet to convert the logfile to a CSV, so I need to wait for that...

RutgerVanBeek (commented):
Changes are fine, but as you said yourself there is an offset and scaling factor, so let's wait until it exactly does what we expect.

PR #276, opened 2020-07-01
update limits after zero position measurement

RutgerVanBeek:
Closes PM-498

## Description

Today we measured the zero positions. We also remeasured all the min and max. 
## Changes

* Changed the min and max position iu.


Roelemans (approved):

Joris-JW (approved):
:fire:

PR #277, opened 2020-07-08
Feature/pm 522 add warning soft limit

Joris-JW:
Closes PM-522

## Description
This PR create a new _error soft limit_ to prevent the exoskeleton from raising an exception when it is just over the soft limits defined in the URDF. It has a new parameter for every joint in the controller.yaml which defines the percentage after which the soft limit should throw an error. When it is inside the _soft_limits_ defined in the URDF but **not** over the _error_soft_limit_ ROS will try and compensate the error and move to an acceptable position.

## Changes
* New error soft limit percentage parameter in controller.yaml for every joint
* Created a soft limit error vector to define the upper and lower error for every joint
* Used the new error soft limit in the outsideLimitsCheck in the hardware interface

### DIFF
--- a/march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
+++ b/march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
@@ -77,8 +77,6 @@ public:
    * Construct the soft limit error based on the margin defined in the controller yaml and the soft limits and hard
    * limits in the URDF.
    */
-  static void getSoftJointLimitsError(const std::string& name, const urdf::JointConstSharedPtr& urdf_joint,
-                                      joint_limits_interface::SoftJointLimits& error_soft_limits);
 
 private:
   void uploadJointNames(ros::NodeHandle& nh) const;
@@ -94,6 +92,8 @@ private:
   void updateIMotionCubeState();
   void outsideLimitsCheck(size_t joint_index);
   bool iMotionCubeStateCheck(size_t joint_index);
+  static void getSoftJointLimitsError(const std::string& name, const urdf::JointConstSharedPtr& urdf_joint,
+                                      joint_limits_interface::SoftJointLimits& error_soft_limits);
 
   /* Limit of the change in effort command over one cycle, can be overridden by safety controller */
   static constexpr double MAX_EFFORT_CHANGE = 5000;
--- a/march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
+++ b/march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
@@ -68,7 +68,17 @@ public:
    */
   int getEthercatCycleTime() const;
 
+  /**
+   * Wait for received PDO.
+   */
   void waitForPdo();
+
+  /**
+   * Construct the soft limit error based on the margin defined in the controller yaml and the soft limits and hard
+   * limits in the URDF.
+   */
+  static void getSoftJointLimitsError(const std::string& name, const urdf::JointConstSharedPtr& urdf_joint,
+                                      joint_limits_interface::SoftJointLimits& error_soft_limits);
 
 private:
   void uploadJointNames(ros::NodeHandle& nh) const;
@@ -120,6 +130,7 @@ private:
   std::vector<double> joint_temperature_variance_;
 
   std::vector<joint_limits_interface::SoftJointLimits> soft_limits_;
+  std::vector<joint_limits_interface::SoftJointLimits> soft_limits_error_;
 
   PowerNetOnOffCommand power_net_on_off_command_;
   bool master_shutdown_allowed_command_ = false;
--- a/march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
+++ b/march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
@@ -73,11 +73,6 @@ public:
    */
   void waitForPdo();
 
-  /**
-   * Construct the soft limit error based on the margin defined in the controller yaml and the soft limits and hard
-   * limits in the URDF.
-   */
-
 private:
   void uploadJointNames(ros::NodeHandle& nh) const;
   /**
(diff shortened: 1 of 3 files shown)


march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -460,20 +470,59 @@ bool MarchHardwareInterface::iMotionCubeStateCheck(size_t joint_index)
...
+  if (!ros::param::has(param_name.str()))
+  {
+    error_stream << "Margin soft limits error of joint: " << name << " could not be found";
+    throw std::runtime_error(error_stream.str());
+  }
+
+  float margin;
+  ros::param::param<float>(param_name.str(), margin, 0.0);
+
+  if (!urdf_joint || !urdf_joint->safety || !urdf_joint->limits || margin == 0.0)
```
Olavhaasie: Why is a margin of `0.0` not allowed?

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -460,20 +470,59 @@ bool MarchHardwareInterface::iMotionCubeStateCheck(size_t joint_index)
...
+  {
+    error_stream << "Could not construct the soft limits for joint: " << name;
+    throw std::runtime_error(error_stream.str());
+  }
+
+  error_soft_limits.min_position =
+      urdf_joint->limits->lower + ((urdf_joint->limits->lower - urdf_joint->safety->soft_lower_limit) * margin);
```
Olavhaasie: I think this should be the other way around.
```suggestion
      urdf_joint->limits->lower + ((urdf_joint->safety->soft_lower_limit - urdf_joint->limits->lower) * margin);
```

march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
```diff
@@ -68,8 +68,18 @@ class MarchHardwareInterface : public hardware_interface::RobotHW
...
+  /**
+   * Wait for received PDO.
+   */
   void waitForPdo();
 
+  /**
+   * Construct the soft limit error based on the margin defined in the controller yaml and the soft limits and hard
+   * limits in the URDF.
+   */
+  static void getSoftJointLimitsError(const std::string& name, const urdf::JointConstSharedPtr& urdf_joint,
```
Olavhaasie: This method can be made private

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -44,15 +44,24 @@ bool MarchHardwareInterface::init(ros::NodeHandle& nh, ros::NodeHandle& /* robot
   this->reserveMemory();
 
   // Start ethercat cycle in the hardware
-  this->march_robot_->startEtherCAT(this->reset_imc_);
+  // this->march_robot_->startEtherCAT(this->reset_imc_);
```
Olavhaasie:
```suggestion
  this->march_robot_->startEtherCAT(this->reset_imc_);
```

march_hardware_interface/src/march_hardware_interface.cpp
```diff
@@ -460,20 +470,59 @@ bool MarchHardwareInterface::iMotionCubeStateCheck(size_t joint_index)
...
+  if (!ros::param::has(param_name.str()))
+  {
+    error_stream << "Margin soft limits error of joint: " << name << " could not be found";
+    throw std::runtime_error(error_stream.str());
+  }
+
+  float margin;
+  ros::param::param<float>(param_name.str(), margin, 0.0);
+
+  if (!urdf_joint || !urdf_joint->safety || !urdf_joint->limits || margin <= 0.0 || margin > 1.0)
```
Olavhaasie, line 518: Why is a margin of 0.0 not allowed?

Olavhaasie (changes requested):

march_hardware_interface/src/march_hardware_interface.cpp
Joris-JW: I guess we would always want to error if we are _too close_ to the hard limits.

march_hardware_interface/src/march_hardware_interface.cpp
Joris-JW: copy error :see_no_evil:

march_hardware_interface/src/march_hardware_interface.cpp
Joris-JW: my bad

march_hardware_interface/src/march_hardware_interface.cpp
Joris-JW: I'll resolve this duplicate question

march_hardware_interface/src/march_hardware_interface.cpp
Olavhaasie: > _too close_

i.e. on the hard limit. Let's see how this works out in practice first :wink:

march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
```diff
@@ -68,8 +68,16 @@ class MarchHardwareInterface : public hardware_interface::RobotHW
...
+  /**
+   * Wait for received PDO.
+   */
   void waitForPdo();
 
+  /**
+   * Construct the soft limit error based on the margin defined in the controller yaml and the soft limits and hard
+   * limits in the URDF.
+   */
+
```
Olavhaasie: oopsie :see_no_evil: you forgot to copy the docs :joy_cat:

Olavhaasie (changes requested):

march_hardware_interface/include/march_hardware_interface/march_hardware_interface.h
Joris-JW: I'm very dumb

march_hardware_interface/src/march_hardware_interface.cpp
Joris-JW: I agee

Roelemans (approved):

Olavhaasie (approved):

PR #278, opened 2020-07-16
Feature/pm 458 implement inertia calculations in a controller

RPKrijnen:
Closes PM-458

## Description

Provides two controllers, an effort controller that might not be used by our exoskeleton, but provides a nice template for M6 to work from and a trajectory controller that has been tested on the exo already and seems to work just like the already implemented one. 
As mentioned, the controller does not yet affect the workings of the exo, but instead provides a custom controller from which an inertia is published. In a follow-up PR I will implement control points and PID interpolation after thoroughly measuring inertias with this PR on a working test-joint.

mooiblokdiagramvangainscheduling2

_How to test:_ If you'd like to test this PR please look at the following branch/PR: https://github.com/project-march/simulation/pull/59 it will implement the controller in simulation so that you can freely test it.

## Changes

* Added inertia_estimator class
* Added inertia controller and its plugin
* Added inertia trajectory controller and its plugin
* Added yaml parameters for easy tuning of the inertia estimator


march_joint_inertia_controller/CMakeLists.txt
```diff
@@ -4,34 +4,53 @@ project(march_joint_inertia_controller)
...
+    include/${PROJECT_NAME}/joint_trajectory_inertia_controller.h
+    src/inertia_estimator.cpp
+    src/joint_inertia_controller.cpp
+    src/joint_trajectory_inertia_controller.cpp)
 target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
 
 install(DIRECTORY include/${PROJECT_NAME}/
     DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
-)
+    )
```
Roelemans: But y

march_joint_inertia_controller/CMakeLists.txt
```diff
@@ -4,34 +4,53 @@ project(march_joint_inertia_controller)
...
 
 install(TARGETS ${PROJECT_NAME}
     ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
     LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-)
+    )
```
Roelemans: But y

march_joint_inertia_controller/CMakeLists.txt
```diff
@@ -4,34 +4,53 @@ project(march_joint_inertia_controller)
...
 
 install(FILES controller_plugins.xml
     DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
-)
+    )
```
Roelemans: But y

march_joint_inertia_controller/CMakeLists.txt
```diff
@@ -4,34 +4,53 @@ project(march_joint_inertia_controller)
...
+    joint_trajectory_controller
+    march_hardware
+    march_shared_resources
     pluginlib
+    realtime_tools
     roscpp
-)
+    std_msgs
+    urdf
+    )
```
Roelemans:
```suggestion
)
```

march_joint_inertia_controller/package.xml
```diff
@@ -1,21 +1,32 @@
 <?xml version="1.0"?>
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
-  <name>march_joint_inertia_controller</name>
-  <version>0.0.0</version>
-  <description>ROS controller which can be customized to operate in different circumstances.  </description>
+    <name>march_joint_inertia_controller</name>
+    <version>0.0.0</version>
+    <description>ROS controller which can be customized to operate in different circumstances.</description>
```
Roelemans: Why the different spacing in this entire file?

march_joint_inertia_controller/config/joint_inertia_controller.yaml
```diff
@@ -1,12 +1,76 @@
...
+            d: [0, 5]
+        linear:
+          std_samples: 100 # number of samples used to calculate the standard deviation
+          lambda: 0.96
+          alpha_filter_size: 12
+          control_points:
+            inertia: [0.0, 1.0]
+            p: [150, 300]
+            d: [0, 5]
+      constraints:
```
Joris-JW, line 70: Should this replace the [controller.yaml](https://github.com/project-march/hardware-interface/blob/develop/march_hardware_interface/config/march4/controllers.yaml)? in that case it should be updated with new values for the softlimits

Joris-JW:
Maybe add a _how to test_ comment in the PR description :computer:

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+    { 2.31330497e-05, 4.62660994e-05, 2.31330497e-05, 1.00000000e+00, -1.37177561e+00, 4.75382129e-01 },
+    { 1.00000000e+00, 2.00000000e+00, 1.00000000e+00, 1.00000000e+00, -1.47548044e+00, 5.86919508e-01 },
+    { 1.00000000e+00, 2.00000000e+00, 1.00000000e+00, 1.00000000e+00, -1.69779140e+00, 8.26021017e-01 }
+  };
+  // namespace joint_inertia_controller
+
+  // Default length 12 for the alpha calculation
+  size_t acc_size_;
+  std::vector<double> acceleration_array_;
+  // Equal to atwo, because two values are needed to calculate the derivative
```
RutgerVanBeek:
```suggestion
  // Equal to two, because two values are needed to calculate the derivative
```

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,211 @@
...
+public:
+  HardwareInterfaceAdapter() : joint_handles_ptr_(nullptr)
+  {
+  }
+
+  /**
+   * \brief Initialize the controller by establishing the pointer to the joints, the pid calculators and the inertia
+   * estimators
+   */
+  bool init(std::vector<hardware_interface::JointHandle>& joint_handles, ros::NodeHandle& nh)
```
RutgerVanBeek, line 44: Shouldn't the implementation be in the .cpp file?

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+{
+  double sum = 0.0;
+  for (size_t i = 0; i < a.size(); ++i)
+  {
+    sum += a[i];
+  }
+  if (sum == 0.0)
+  {
+    return 0.0;
+  }
```
RutgerVanBeek: This check does not add anything, right? It might be a bit faster in an edge case, but is it worth it? If you do this, because you do not want to divide by zero, you should check whether the size is zero.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+  filtered_acceleration_array_.resize(acc_size_);
+
+  joint_torque_.resize(torque_size_);
+
+  filtered_joint_torque_.resize(fil_tor_size_);
+
+  for (unsigned int i = 0; i < vel_size_; ++i)
+  {
+    velocity_array_[i] = 0.0;
```
RutgerVanBeek: Doesn't it default to zero?

RutgerVanBeek (changes requested):
This was a partial review, but may be you can already react to this.

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
RPKrijnen: No, because this is a template the functionality should be added right here in the header file.

march_joint_inertia_controller/src/inertia_estimator.cpp
RPKrijnen: you are correct, I will change this

march_joint_inertia_controller/src/inertia_estimator.cpp
RPKrijnen: I wasn't sure, if it does it does remove some unnecessary lines, which would be nice.
edit: You were right, it does not break the simulation and I will therefore implement your suggestion :)

march_joint_inertia_controller/config/joint_inertia_controller.yaml
RPKrijnen: I will update them

march_joint_inertia_controller/CMakeLists.txt
RPKrijnen: Cntrl+Alt+L

march_joint_inertia_controller/CMakeLists.txt
RPKrijnen: Cntrl+Alt+L

march_joint_inertia_controller/CMakeLists.txt
RPKrijnen: Cntrl+Alt+L

march_joint_inertia_controller/package.xml
RPKrijnen: probably because of: Cntrl+Alt+L

Deakensbubbles: Cntrl+Alt+L

march_joint_inertia_controller/CMakeLists.txt
Olavhaasie: I don't think this is auto formatting, since some brackets are indented and some are not, Could you please these changes :innocent:

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+#include <trajectory_interface/quintic_spline_segment.h>
+#include <urdf/model.h>
+
+/**
+ * \brief determines the absolute value of vector a and stores it into b so that a remains unchanged
+ *
+ * @param[in] a input vector with values that need to be converted to absolutes
+ * @param[out] b output vector with copies of the absolute values of a.
+ */
+void absolute(std::vector<double> a, std::vector<double>& b);
```
Olavhaasie: Why not just return the vector? Then you don't have to initialize a vector first before calling this and this is what return values are for. Furthermore, it is best to pass a vector as a const reference if you don't want to modify it, since it does not require a copy.
```suggestion
std:vector<double> absolute(const std::vector<double>& a);
```

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+  hardware_interface::JointHandle joint_;
+
+  float min_alpha_ = 0.4;  // You might want to be able to adjust this value from a yaml/launch file
+  float max_alpha_ = 0.9;  // You might want to be able to adjust this value from a yaml/launch file
+
+  double sos_[3][6] = {
+    { 2.31330497e-05, 4.62660994e-05, 2.31330497e-05, 1.00000000e+00, -1.37177561e+00, 4.75382129e-01 },
+    { 1.00000000e+00, 2.00000000e+00, 1.00000000e+00, 1.00000000e+00, -1.47548044e+00, 5.86919508e-01 },
+    { 1.00000000e+00, 2.00000000e+00, 1.00000000e+00, 1.00000000e+00, -1.69779140e+00, 8.26021017e-01 }
+  };
```
Olavhaasie: What are these numbers? Perhaps a comment would be helpful or a more descriptive name.

march_joint_inertia_controller/package.xml
Olavhaasie: Could you please revert this change to 2 spaces, it is what we use in every package manifest.

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+ */
+void absolute(std::vector<double> a, std::vector<double>& b);
+
+/**
+ * \brief determiness the mean value of the given vector
+ *
+ * @param[in] a input vector from which the mean gets calculated
+ * @returns the mean value of a
+ */
+double mean(std::vector<double> a);
```
Olavhaasie: Same here, pass the vector as a const reference
```suggestion
double mean(const std::vector<double>& a);
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  }
+}
+
+double mean(std::vector<double> a)
+{
+  double sum = 0.0;
+  for (size_t i = 0; i < a.size(); ++i)
+  {
+    sum += a[i];
+  }
```
Olavhaasie:
```suggestion
  for (const double x : a)
  {
    sum += x;
  }
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  // Size the buffers
+  velocity_array_.resize(vel_size_);
+
+  acceleration_array_.resize(acc_size_);
+
+  filtered_acceleration_array_.resize(acc_size_);
+
+  joint_torque_.resize(torque_size_);
+
+  filtered_joint_torque_.resize(fil_tor_size_);
```
Olavhaasie: To be completely sure they are all initialized to 0
```suggestion
  velocity_array_.resize(vel_size_, 0.0);

  acceleration_array_.resize(acc_size_, 0.0);

  filtered_acceleration_array_.resize(acc_size_, 0.0);

  joint_torque_.resize(torque_size_, 0.0);

  filtered_joint_torque_.resize(fil_tor_size_, 0.0);
```

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * \brief Constructor that accepts a joint
+   */
+  InertiaEstimator(hardware_interface::JointHandle joint, double lambda = 0.96, size_t acc_size = 12);
+
+  /**
+   * \brief Destructor
+   */
+  ~InertiaEstimator();
```
Olavhaasie: This destructor is not necessary, since you leave it empty.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  {
+    acceleration_array_[i] = 0.0;
+  }
+
+  for (unsigned int i = 0; i < torque_size_; ++i)
+  {
+    joint_torque_[i] = 0.0;
+  }
+
+  // Setup Butterworth filter
```
Olavhaasie: Is this a TODO? If so, why didn't you do it?

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
+// Copyright 2020 Project March.
+#include <hardware_interface/joint_command_interface.h>
+#include <hardware_interface/joint_state_interface.h>
+#include "march_joint_inertia_controller/inertia_estimator.h"
+#include "math.h"
```
Olavhaasie: `math.h` is an old deprecated header, you should use `cmath`
```suggestion
#include <cmath>
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+    acceleration_array_[i] = 0.0;
+  }
+
+  for (unsigned int i = 0; i < torque_size_; ++i)
+  {
+    joint_torque_[i] = 0.0;
+  }
+
+  // Setup Butterworth filter
+}
```
Olavhaasie: You can actually call the first constructor in the initializer list of this one, the only extra call is `setJoint`.
```suggestion
  : InertiaEstimator(lambda, acc_size)
{
  setJoint(joint);
}
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  it = joint_torque_.begin();
+  it = joint_torque_.insert(it, effort);
+  joint_torque_.resize(torque_size_);
+  return false;
+}
+
+void InertiaEstimator::apply_butter()
+{
+  // Dingen doen met de sos filter ofzo
+  // Also doe pushback dingen anders dan werkt dit niet
```
Olavhaasie: English please :gb:

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+{
+  acc_size_ = acc_size;
+}
+
+void InertiaEstimator::setNodeHandle(ros::NodeHandle& nh)
+{
+  nh_ = nh;
+}
+
+void InertiaEstimator::configurePublisher(std::string name)
```
Olavhaasie: use const reference to avoid copies
```suggestion
void InertiaEstimator::configurePublisher(const std::string& name)
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+bool InertiaEstimator::fill_buffers(double velocity, double effort, const ros::Duration& period)
+{
+  auto it = velocity_array_.begin();
+
+  it = velocity_array_.begin();
+  it = velocity_array_.insert(it, velocity);
+  velocity_array_.resize(vel_size_);
+
+  // Automatically fills the zero'th position of the acceleration array
+  discrete_speed_derivative(velocity, period);
```
Olavhaasie: I think it would be a lot easier to read if this method returned its calculated derivative instead of adding it as a side effect. Now I have to look up the method to see what it does. Use the return values to create mostly 'pure' functions without side effects.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// Fills the buffers so that non-zero values may be computed by the inertia estimator
+bool InertiaEstimator::fill_buffers(double velocity, double effort, const ros::Duration& period)
+{
+  auto it = velocity_array_.begin();
+
+  it = velocity_array_.begin();
+  it = velocity_array_.insert(it, velocity);
+  velocity_array_.resize(vel_size_);
```
Olavhaasie: This piece of code is kinda hard to understand. I think what you're doing is inserting an element in the front and removing the last one? Kinda like a ring buffer. Keep in mind that [`insert`](http://www.cplusplus.com/reference/vector/vector/insert/) has to move all elements in the array one up, so this is a very ineffecient implementation of a ring buffer. It requires a new allocation every time, which could greatly reduce performance and determinism in a real time application like a controller. You could use a [`list`](http://www.cplusplus.com/reference/list/list/) with the downside being that you can't random access elements. If you do need the random access, take a look at ringbuffer implementations: https://github.com/lava/linear_ringbuffer, https://github.com/idreamerhx/ringbuffer

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+  // Automatically fills the zero'th position of the acceleration array
+  discrete_speed_derivative(velocity, period);
+  acceleration_array_.resize(acc_size_);
+
+  it = joint_torque_.begin();
+  it = joint_torque_.insert(it, effort);
+  joint_torque_.resize(torque_size_);
```
Olavhaasie: Same for this inserting and resizing here.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  return false;
```
Olavhaasie: This method always returns `false`?

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  // Dingen doen met de sos filter ofzo
+  // Also doe pushback dingen anders dan werkt dit niet
+  std::vector<double> z = { 0.0, 0.0, 0.0 };
+  auto it = filtered_acceleration_array_.begin();
+  auto x_n = acceleration_array_[0];
+  double x;
+
+  for (unsigned int i = 0; i < (sizeof(sos_) / sizeof(sos_[0])); ++i)
+  {
+    x_n = acceleration_array_[0];
```
Olavhaasie: This line does nothing? It keeps setting `x_n` to the same value.

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * \brief Get the position of the joint stored in the private variable
+   */
+  double getPosition();
```
Olavhaasie: This method is not implemented?

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+  void setLambda(double lambda);
+  void setAcc_size(size_t acc_size);
+  void setNodeHandle(ros::NodeHandle& nh);
+  /**
+   * \brief Initialize the publisher with a name
+   */
+  void configurePublisher(std::string name);
+  void publishInertia();
+
+  std::string getJointName();
```
Olavhaasie: This method is also not implemented

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+private:
+  ros::NodeHandle nh_;
+  ros::Publisher pub_;
+
+  hardware_interface::JointHandle joint_;
+
+  float min_alpha_ = 0.4;  // You might want to be able to adjust this value from a yaml/launch file
+  float max_alpha_ = 0.9;  // You might want to be able to adjust this value from a yaml/launch file
+
+  double sos_[3][6] = {
```
Olavhaasie: It would be nice to create constants of 3 and 6, so that you can use them in the for loop, without having to do the complicated `sizeof` stuff. Furthermore, this variable can be made `static const`, since they are always the same.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  it = joint_torque_.insert(it, effort);
+  joint_torque_.resize(torque_size_);
+  return false;
+}
+
+void InertiaEstimator::apply_butter()
+{
+  // Dingen doen met de sos filter ofzo
+  // Also doe pushback dingen anders dan werkt dit niet
+  std::vector<double> z = { 0.0, 0.0, 0.0 };
```
Olavhaasie: `z[2]` is never used?

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  {
+    x_n = acceleration_array_[0];
+    x = sos_[i][0] * x_n + z[0];
+    z[0] = sos_[i][1] * x_n - sos_[i][3] * x + z[1];
+    z[1] = sos_[i][2] * x_n - sos_[i][4] * x;
+
+    //    temp *= (acceleration_array_[0] * sos_[i][0] + acceleration_array_[1] * sos_[i][1] +
+    //             acceleration_array_[2] * sos_[i][2]) /
+    //            (acceleration_array_[0] * sos_[i][3] + acceleration_array_[1] * sos_[i][4] +
+    //             acceleration_array_[2] * sos_[i][5]);
```
Olavhaasie: Why is this commented? Commented code should be removed, before it stays there forever.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+void InertiaEstimator::apply_butter()
+{
+  // Dingen doen met de sos filter ofzo
+  // Also doe pushback dingen anders dan werkt dit niet
+  std::vector<double> z = { 0.0, 0.0, 0.0 };
+  auto it = filtered_acceleration_array_.begin();
+  auto x_n = acceleration_array_[0];
+  double x;
+
+  for (unsigned int i = 0; i < (sizeof(sos_) / sizeof(sos_[0])); ++i)
```
Olavhaasie: Use range based for loops
```suggestion
  for (const auto& so : sos_)
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+    //             acceleration_array_[2] * sos_[i][2]) /
+    //            (acceleration_array_[0] * sos_[i][3] + acceleration_array_[1] * sos_[i][4] +
+    //             acceleration_array_[2] * sos_[i][5]);
+  }
+  it = filtered_acceleration_array_.begin();
+  it = filtered_acceleration_array_.insert(it, x);
+  filtered_acceleration_array_.resize(acc_size_);
+
+  z = { 0.0, 0.0, 0.0 };
+  for (unsigned int i = 0; i < (sizeof(sos_) / sizeof(sos_[0])); ++i)
```
Olavhaasie:
```suggestion
  for (const auto& so : sos_)
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+    z[0] = sos_[i][1] * x_n - sos_[i][3] * x + z[1];
+    z[1] = sos_[i][2] * x_n - sos_[i][4] * x;
+    //    temp *= (joint_torque_[0] * sos_[i][0] + joint_torque_[1] * sos_[i][1] + joint_torque_[2] * sos_[i][2]) /
+    //            (joint_torque_[0] * sos_[i][3] + joint_torque_[1] * sos_[i][4] + joint_torque_[2] * sos_[i][5]);
+  }
+  it = filtered_joint_torque_.begin();
+  it = filtered_joint_torque_.insert(it, x);
+  filtered_joint_torque_.resize(fil_tor_size_);
+
+  // temp = 1.0;
```
Olavhaasie: Please remove commented out code

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+  z = { 0.0, 0.0, 0.0 };
+  for (unsigned int i = 0; i < (sizeof(sos_) / sizeof(sos_[0])); ++i)
+  {
+    x_n = joint_torque_[0];
+    x = sos_[i][0] * x_n + z[0];
+    z[0] = sos_[i][1] * x_n - sos_[i][3] * x + z[1];
+    z[1] = sos_[i][2] * x_n - sos_[i][4] * x;
+    //    temp *= (joint_torque_[0] * sos_[i][0] + joint_torque_[1] * sos_[i][1] + joint_torque_[2] * sos_[i][2]) /
+    //            (joint_torque_[0] * sos_[i][3] + joint_torque_[1] * sos_[i][4] + joint_torque_[2] * sos_[i][5]);
```
Olavhaasie: Please remove commented out code

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  filtered_joint_torque_.resize(fil_tor_size_);
+
+  // temp = 1.0;
+}
+
+// Estimate the inertia using the acceleration and torque
+void InertiaEstimator::inertia_estimate()
+{
+  double torque_e;
+  double acc_e;
```
Olavhaasie: You don't have to initialize variables at the start of a method, instead just initaliaze them when you need them

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  apply_butter();
+
+  correlation_calculation();
+  K_i_ = gain_calculation();
+  K_a_ = alpha_calculation();
+  torque_e = filtered_joint_torque_[0] - filtered_joint_torque_[1];
+  acc_e = filtered_acceleration_array_[0] - filtered_acceleration_array_[1];
```
Olavhaasie:
```suggestion
  const double torque_e = filtered_joint_torque_[0] - filtered_joint_torque_[1];
  const double acc_e = filtered_acceleration_array_[0] - filtered_acceleration_array_[1];
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+{
+  double vib = vibration_calculation();
+  if (vib < min_alpha_)
+  {
+    vib = min_alpha_;
+  }
+  else if (vib > max_alpha_)
+  {
+    vib = max_alpha_;
+  }
```
Olavhaasie:
```suggestion
  double vib = std::max(std::min(vibration_calculation(), min_alpha_), max_alpha_);
```

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+  std::vector<double> standard_deviation;
+
+private:
+  ros::NodeHandle nh_;
+  ros::Publisher pub_;
+
+  hardware_interface::JointHandle joint_;
+
+  float min_alpha_ = 0.4;  // You might want to be able to adjust this value from a yaml/launch file
+  float max_alpha_ = 0.9;  // You might want to be able to adjust this value from a yaml/launch file
```
Olavhaasie:
```suggestion
  double min_alpha_ = 0.4;  // You might want to be able to adjust this value from a yaml/launch file
  double max_alpha_ = 0.9;  // You might want to be able to adjust this value from a yaml/launch file
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+  return (corr_coeff_ * (filtered_acceleration_array_[0] - filtered_acceleration_array_[1])) /
+         (lambda_ + corr_coeff_ * pow(filtered_acceleration_array_[0] - filtered_acceleration_array_[1], 2));
+}
+
+// Calculate the correlation coefficient of the acceleration buffer
+void InertiaEstimator::correlation_calculation()
+{
+  corr_coeff_ =
+      corr_coeff_ / (lambda_ + corr_coeff_ * pow(filtered_acceleration_array_[0] - filtered_acceleration_array_[1], 2));
+  double large_number = pow(10, 8);
```
Olavhaasie: computing the power takes time, so you can convert this to a constant.
```suggestion
  const double large_number = 10e8;
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+double InertiaEstimator::vibration_calculation()
+{
+  std::vector<double> b;
+  b.resize(acc_size_);
+  absolute(filtered_acceleration_array_, b);
+  // moa = mean of the absolute
+  moa_ = mean(b);
+  // aom = absolute of the mean
+  aom_ = abs(mean(filtered_acceleration_array_));
+  // Divide by zero protection, necessary when exo hasn't homed yet
```
Olavhaasie:
```suggestion
  // Divide by zero protection, necessary when there has not been any acceleration yet
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+{
+  // Setup the initial value for the correlation coefficient 100*standarddeviation(acceleration)^2
+  double mean_value = mean(standard_deviation);
+  double sum = 0;
+  for (unsigned int i = 0; i < standard_deviation.size(); ++i)
+  {
+    sum += pow((standard_deviation[i] - mean_value), 2);
+  }
+  corr_coeff_ = 100 * (sum / samples);
+  return;
```
Olavhaasie: `return` is not necessary for `void` methods.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,290 @@
...
+}
+
+void InertiaEstimator::init_p(unsigned int samples)
+{
+  // Setup the initial value for the correlation coefficient 100*standarddeviation(acceleration)^2
+  double mean_value = mean(standard_deviation);
+  double sum = 0;
+  for (unsigned int i = 0; i < standard_deviation.size(); ++i)
+  {
+    sum += pow((standard_deviation[i] - mean_value), 2);
```
Olavhaasie:
```suggestion
  for (const auto& deviation : standard_deviation)
  {
    sum += std::pow(deviation - mean_value, 2);
```

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+
+  void setJoint(hardware_interface::JointHandle joint);
+  void setLambda(double lambda);
+  void setAcc_size(size_t acc_size);
+  void setNodeHandle(ros::NodeHandle& nh);
+  /**
+   * \brief Initialize the publisher with a name
+   */
+  void configurePublisher(std::string name);
+  void publishInertia();
```
Olavhaasie: I think it would be better to remove this publishing functionality from the inertia estimator and add it to the controller instead. Right now this requires all kind of extra methods to bloat this class, which clearly has a single function. Whereas the controller already has publishers and a node handle.

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+  void apply_butter();
+
+  /**
+   * \brief Estimate the inertia using the acceleration and torque
+   */
+  void inertia_estimate();
+  /**
+   * \brief Calculate a discrete derivative of the speed measurements
+   */
+  void discrete_speed_derivative(double velocity, const ros::Duration&);
```
Olavhaasie: Please give a name to the second parameter

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+   */
+  bool fill_buffers(double velocity, double effort, const ros::Duration& period);
+
+  /**
+   * \brief Calculate the initial correlation coefficient
+   */
+  void init_p(unsigned int samples);
+
+  urdf::JointConstSharedPtr joint_urdf_;
+  std::string joint_name;
```
Olavhaasie: These members are not used?

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,157 @@
...
+  // Vector to be filled with samples of acceleration to determine the standard deviation from
+  std::vector<double> standard_deviation;
+
+private:
+  ros::NodeHandle nh_;
+  ros::Publisher pub_;
+
+  hardware_interface::JointHandle joint_;
```
Olavhaasie: This one is also not used

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,211 @@
...
+  {
+    joint_handles_ptr_ = &joint_handles;
+    const unsigned int num_joints_ = joint_handles_ptr_->size();
```
Olavhaasie: You shadow the private member `num_joints_` here. So, you either meant to use that one or you should rename it

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,211 @@
...
+
+  /**
+   * \brief Updates the commanded effort for each individual joint and estimates the inertia on each joint and publishes
+   * that information on a topic
+   */
+  void updateCommand(const ros::Time& /*time*/, const ros::Duration& period,
+                     const joint_trajectory_controller::State& /*desired state*/,
+                     const joint_trajectory_controller::State& state_error)
+  {
+    const unsigned int num_joints_ = joint_handles_ptr_->size();
```
Olavhaasie: Again you shadow the member here

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,211 @@
...
+
+    // Preconditions
+    if (!joint_handles_ptr_)
+      return;
```
Olavhaasie: We like to use brackets here
```suggestion
    if (!joint_handles_ptr_)
    {
      return;
    }
```

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,211 @@
...
+
+namespace joint_trajectory_controller
+{
+typedef trajectory_interface::QuinticSplineSegment<double> SegmentImpl;
+typedef JointTrajectorySegment<SegmentImpl> Segment;
+typedef typename Segment::State State;
+}  // namespace joint_trajectory_controller
+
+template <>
+class HardwareInterfaceAdapter<hardware_interface::EffortJointInterface, joint_trajectory_controller::State>
```
Olavhaasie, line 33: I don't fully understand what this class does and what the difference is between the `IntertiaController`. As far as I can tell, this "controller" does not implement the controller interface like `IntertiaController` does. And why does it use an empty template?

march_joint_inertia_controller/src/joint_trajectory_inertia_controller.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+namespace inertia_trajectory_controller
+{
+typedef joint_trajectory_controller::JointTrajectoryController<trajectory_interface::QuinticSplineSegment<double>,
+                                                               hardware_interface::EffortJointInterface>
+    JointTrajectoryController;
+
+}  // namespace inertia_trajectory_controller
+
+PLUGINLIB_EXPORT_CLASS(inertia_trajectory_controller::JointTrajectoryController, controller_interface::ControllerBase);
```
Olavhaasie, line 13: So the `intertia_trajectory_controller` is just exactly the same as the effort controller? Since you export it as the `EffortJointInterface?

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,211 @@
...
+  /**
+   * \brief Set the PID parameters
+   */
+  void setGains(int joint_nr, const double& p, const double& i, const double& d, const double& i_max,
+                const double& i_min, const bool& antiwindup = false)
+  {
+    (*pids_[joint_nr]).setGains(p, i, d, i_max, i_min, antiwindup);
+  }
+
+  std::string joint_names;
```
Olavhaasie: This member is not used.

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,211 @@
...
+  int samples_;
+
+  typedef std::shared_ptr<control_toolbox::Pid> PidPtr;
+  std::vector<PidPtr> pids_;
+
+  std::vector<hardware_interface::JointHandle>* joint_handles_ptr_;
+
+  unsigned int num_joints_;
+  std::vector<hardware_interface::JointHandle> joints_;
+  double init_pos_;
```
Olavhaasie: These 3 are also never used

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
```diff
@@ -3,23 +3,82 @@
...
 {
 public:
+  struct Commands
+  {
+    double position_;    // Last commanded position
+    double velocity_;    // Last commanded velocity
+    bool has_velocity_;  // false if no velocity command has been specified
+  };
+
+  InertiaController();
```
Olavhaasie: You have an empty constructor, so you can use `default` and remove the implementation
```suggestion
  InertiaController() = default;
```

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
```diff
@@ -3,23 +3,82 @@
...
 {
-class InertiaController : public controller_interface::Controller<hardware_interface::PositionJointInterface>
+class InertiaController : public controller_interface::Controller<hardware_interface::EffortJointInterface>
 {
 public:
+  struct Commands
+  {
+    double position_;    // Last commanded position
+    double velocity_;    // Last commanded velocity
+    bool has_velocity_;  // false if no velocity command has been specified
```
Olavhaasie: These variables are not private and should therefore not have the underscore

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
```diff
@@ -3,23 +3,82 @@
...
+  };
+
+  InertiaController();
+  ~InertiaController();
```
Olavhaasie: You shoudl add `override` to methods you override from the base class
```suggestion
  ~InertiaController() override;
```

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
```diff
@@ -3,23 +3,82 @@
...
+
   bool init(hardware_interface::PositionJointInterface* hw, ros::NodeHandle& n);
   void update(const ros::Time& time, const ros::Duration& period);
   void starting(const ros::Time& time);
   void stopping(const ros::Time& time);
```
Olavhaasie: I know this was already merged, but it is still important
```suggestion
  bool init(hardware_interface::PositionJointInterface* hw, ros::NodeHandle& n) override;
  void update(const ros::Time& time, const ros::Duration& period) override;
  void starting(const ros::Time& time) override;
  void stopping(const ros::Time& time) override;
```

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
```diff
@@ -3,23 +3,82 @@
...
+   */
+  std::string getJointName();
+
+  urdf::JointConstSharedPtr joint_urdf_;
+  std::string joint_name;
+  realtime_tools::RealtimeBuffer<Commands> command_;
+  Commands command_struct_;  // pre-allocated memory that is re-used to set the realtime buffer
 
 private:
+  int loop_count_;
```
Olavhaasie: Never used

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
```diff
@@ -3,23 +3,82 @@
...
+  control_toolbox::Pid pid_controller_; /**< Internal PID controller. */
+
+  ros::Subscriber sub_command_;
   hardware_interface::JointHandle joint_;
   double init_pos_;
```
Olavhaasie: Never used

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
```diff
@@ -3,23 +3,82 @@
...
+
+  InertiaEstimator inertia_estimator_;
+  /**
+   * \brief Callback from /command subscriber for setpoint
+   */
+  void setCommandCB(const std_msgs::Float64ConstPtr& msg);
```
Olavhaasie: This is not implemented

march_joint_inertia_controller/src/joint_inertia_controller.cpp
```diff
@@ -1,34 +1,182 @@
 // Copyright 2020 Project March.
 #include "march_joint_inertia_controller/joint_inertia_controller.h"
+#include <control_msgs/FollowJointTrajectoryAction.h>
+#include <trajectory_msgs/JointTrajectory.h>
 #include <math.h>
```
Olavhaasie:
```suggestion
#include <cmath>
```

march_joint_inertia_controller/src/joint_inertia_controller.cpp
```diff
@@ -1,34 +1,182 @@
 // Copyright 2020 Project March.
 #include "march_joint_inertia_controller/joint_inertia_controller.h"
+#include <control_msgs/FollowJointTrajectoryAction.h>
```
Olavhaasie: This `include` is never used

march_joint_inertia_controller/src/joint_inertia_controller.cpp
```diff
@@ -1,34 +1,182 @@
...
+  command_struct_.position_ = pos_command;
+  command_struct_.has_velocity_ = false;
+
+  command_.initRT(command_struct_);
+
+  pid_controller_.reset();
 }
 
-void InertiaController::update(const ros::Time& /* time */, const ros::Duration& /* period */)
+void InertiaController::update(const ros::Time& /* time */, const ros::Duration& period /* period */)
```
Olavhaasie:
```suggestion
void InertiaController::update(const ros::Time& /* time */, const ros::Duration& period)
```

march_joint_inertia_controller/src/joint_trajectory_inertia_controller.cpp
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 Project March.
+#include <angles/angles.h>
+#include <control_msgs/FollowJointTrajectoryAction.h>
+#include "march_joint_inertia_controller/joint_trajectory_inertia_controller.h"
+#include <math.h>
+#include <pluginlib/class_list_macros.h>
+#include <pluginlib/class_list_macros.hpp>
+#include <trajectory_msgs/JointTrajectory.h>
```
Olavhaasie:
```suggestion
#include "march_joint_inertia_controller/joint_trajectory_inertia_controller.h"
#include <pluginlib/class_list_macros.hpp>
```

Olavhaasie (changes requested):
I think you did a nice job on making this work on the exo, but there is still room for improvement :+1:

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
RPKrijnen: The main difference between the controllers is that this one accepts trajectory commands, whereas the other needs a specified reference value meaning; this one is the same as the one used on the exo already. The other is a nice to have as far as I see it. I have put some work into fixing it before realizing it was not what we directly needed.

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
RPKrijnen: I just tested this, but the compiler seems to think it doesn't override the init function.

march_joint_inertia_controller/src/joint_trajectory_inertia_controller.cpp
RPKrijnen: not exactly, the main difference lies in that the trajectory controller gets fed its reference values by the FollowJointTrajectory actions. these are then given through to the controller. The effort controller can have any self defined interface i.e. not reliant on a gait message.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -152,23 +152,15 @@ void InertiaEstimator::inertia_estimate()
...
-  double vib = vibration_calculation();
-  if (vib < min_alpha_)
-  {
-    vib = min_alpha_;
-  }
-  else if (vib > max_alpha_)
-  {
-    vib = max_alpha_;
-  }
+  double vib = std::max(std::min(vibration_calculation(), min_alpha_), max_alpha_);
```
RutgerVanBeek: I think you want it the other way around with the min and max but what you really want is std::clamp.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -198,7 +190,7 @@ double InertiaEstimator::vibration_calculation()
   moa_ = mean(absolute(filtered_acceleration_array_));
   // aom = absolute of the mean
   aom_ = abs(mean(filtered_acceleration_array_));
```
RutgerVanBeek: Instead of adding the comment, why not rename the variable to a more descriptive name like absolute_of_mean ?

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,209 @@
...
+        return false;
+      }
+    }
+
+    // Obtain inertia estimator parameters from server
+    double lambda[2];
+    int alpha_filter_size[2];
+    ros::NodeHandle rotary_estimator_nh(nh, std::string("inertia_estimator/rotary"));
+    rotary_estimator_nh.param("std_samples", samples_, 100);
+    rotary_estimator_nh.param("lambda", lambda[0], 0.96);
```
RutgerVanBeek: 0.96, is a bit of an arbitrary default value, may be choose one, using only the last value, seems like a safe choice.

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,209 @@
...
+    {
+      for (unsigned int i = 0; i < num_joints_; ++i)
+      {
+        // Update inertia estimator
+        inertia_estimators_[i].fill_buffers((*joint_handles_ptr_)[i].getVelocity(),
+                                            (*joint_handles_ptr_)[i].getEffort(), period);
+        inertia_estimators_[i].inertia_estimate();
+        inertia_estimators_[i].publishInertia();
+        // TO DO: Provide lookup table for gain selection
+        // TO DO: apply PID control
```
RutgerVanBeek, line 143: I couldn't find Jira issues for this, is that correct? If so I think you should add them, so they are not forgotten.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,219 @@
...
+  double sum = 0.0;
+  for (const double x : a)
+  {
+    sum += x;
+  }
+  if (a.size() == 0.0)
+  {
+    return 0.0;
+  }
+  return sum / a.size();
```
RutgerVanBeek: I just thought if this now, but I you don't want to have the edge case, you can do as follows. It is up to you what you think is better code.
```suggestion
  for (const double x : a)
  {
    mean+= x / a.size();
  }
  return mean;
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,219 @@
...
+  }
+}
+
+// Calculate the vibration based on the acceleration
+double InertiaEstimator::vibration_calculation()
+{
+  // moa = mean of the absolute
+  moa_ = mean(absolute(filtered_acceleration_array_));
+  // aom = absolute of the mean
+  aom_ = abs(mean(filtered_acceleration_array_));
```
RutgerVanBeek: I think it is nice to remove the comments and have a more descriptive variable name as you suggest yourself in the comments.

march_joint_inertia_controller/src/joint_inertia_controller.cpp
```diff
@@ -1,34 +1,175 @@
...
+    // Set the PID error and compute the PID command with nonuniform
+    // time step size.
+    commanded_effort = pid_controller_.computeCommand(error, period);
+  }
+
+  inertia_estimator_.fill_buffers(joint_.getVelocity(), joint_.getEffort(), period);
+  inertia_estimator_.inertia_estimate();
+  inertia_estimator_.publishInertia();
+  // TO DO: Provide lookup table for gain selection
+  // TO DO: apply PID control
```
RutgerVanBeek, line 139: I did not find JIRA issues for this. I think you should add them, because otherwise they will be forgotten.

RutgerVanBeek (changes requested):
Sorry for finishing the review only now, still some commands left.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+double mean(const std::vector<double>& a)
+{
+  double sum = 0.0;
+  for (const double x : a)
+  {
+    sum += x;
+  }
+  if (a.size() == 0.0)
```
Olavhaasie: Size is an integer
```suggestion
  if (a.size() == 0)
```

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,145 @@
...
+public:
+  /**
+   * \brief Default constructor
+   */
+  InertiaEstimator(double lambda = 0.96, size_t acc_size = 12);
+
+  double getAcceleration(unsigned int index);
+
+  void setLambda(double lambda);
+  void setAcc_size(size_t acc_size);
```
Olavhaasie:
```suggestion
  void setAccSize(size_t acc_size);
```

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+   * \brief Fill the rolling buffers with corresponding values.
+   */
+  void fill_buffers(double velocity, double effort, const ros::Duration& period);
+
+  /**
+   * \brief Calculate the initial correlation coefficient
+   */
+  void init_p(unsigned int samples);
```
Olavhaasie: I only noticed this now, but we like to use camelCase naming for class methods. You even use them in this class as well.

Olavhaasie (changes requested):

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
Olavhaasie: That is because this `init` still accepts the `PositionJointInterface` and not the `EffortJointInterface`.

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
RPKrijnen: Done!

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
RPKrijnen: yeah, that's a fine suggestion. I'll change it.

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
RPKrijnen: That was correct, I have added them recently

march_joint_inertia_controller/src/joint_inertia_controller.cpp
RPKrijnen: see my other comment

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
RPKrijnen: I have added a nice comment describing the second order sections representation of the filter

march_joint_inertia_controller/src/inertia_estimator.cpp
RPKrijnen: I think I'll stick with the edge case for now, but thank you for the suggestion

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
RPKrijnen: fixed it

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
```diff
@@ -3,23 +3,76 @@
...
+
+  /**
+   * \brief Return the name of the corresponding joint is string format
+   */
+  std::string getJointName();
+
+  urdf::JointConstSharedPtr joint_urdf_;
+  std::string joint_name;
+  realtime_tools::RealtimeBuffer<Commands> command_;
+  Commands command_struct_;  // pre-allocated memory that is re-used to set the realtime buffer
```
Olavhaasie: Why are these members public?

march_joint_inertia_controller/CMakeLists.txt
```diff
@@ -1,26 +1,45 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(march_joint_inertia_controller)
 
-add_compile_options(-std=c++14 -Wall -Wextra -Werror)
+add_compile_options(-std=c++17 -Wall -Wextra -Werror)
```
Olavhaasie: Why did you upgrade to `c++17`. I think ROS Melodic only supports `c++14`

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
```diff
@@ -3,23 +3,76 @@
...
+  {
+    double position;    // Last commanded position
+    double velocity;    // Last commanded velocity
+    bool has_velocity;  // false if no velocity command has been specified
+  };
+
+  InertiaController() = default;
+  ~InertiaController() override;
+
   bool init(hardware_interface::PositionJointInterface* hw, ros::NodeHandle& n);
```
Olavhaasie: I think this should fix it so that you can override.
```suggestion
  bool init(hardware_interface::EffortJointInterface* hw, ros::NodeHandle& n) override;
```

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,214 @@
...
+    pub_.resize(num_joints_);
+
+    // Initialize PIDs
+    pids_.resize(num_joints_);
+    for (unsigned int i = 0; i < pids_.size(); ++i)
+    {
+      ros::NodeHandle joint_nh(nh, std::string("gains/") + joint_handles[i].getName());
+
+      // Init PID gains from ROS parameter server
+      pids_[i].reset(new control_toolbox::Pid());
```
Olavhaasie: Try to avoid the use of `new` by using `c++14` features for smart pointers.
```suggestion
      pids_[i] = std::make_shared<control_toolbox::Pid>();
```

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
```diff
@@ -0,0 +1,214 @@
...
+  int samples_;
+
+  typedef std::shared_ptr<control_toolbox::Pid> PidPtr;
+  std::vector<PidPtr> pids_;
+
+  std::vector<hardware_interface::JointHandle>* joint_handles_ptr_;
+
+  unsigned int num_joints_;
+
+  std::vector<ros::Publisher> pub_;
```
Olavhaasie: You should make these publishers `realtime_tools::RealtimePublisher`. Because you publish inside of your update loop, which should be real time, these publishers can do that in real time. Take a look at how we use them in `march_hardware_interface.cpp`. You have even already included them at the top of this file, but you are not using them.

Olavhaasie (changes requested):
I went over your code another time, to see if you implemented the suggestions correctly and checked if I missed anything else (since it is quite big :b:)

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -42,83 +42,83 @@ class InertiaEstimator
...
+  double min_alpha_ = 0.4;  // You might want to be able to adjust this value from a yaml/launch file
+  double max_alpha_ = 0.9;  // You might want to be able to adjust this value from a yaml/launch file
 
   // This is a sixth order butterworth filter with a cutoff frequency at 15Hz in Second Order Sections form
-  double sos_[3][6] = {
-    { 2.31330497e-05, 4.62660994e-05, 2.31330497e-05, 1.00000000e+00, -1.37177561e+00, 4.75382129e-01 },
-    { 1.00000000e+00, 2.00000000e+00, 1.00000000e+00, 1.00000000e+00, -1.47548044e+00, 5.86919508e-01 },
-    { 1.00000000e+00, 2.00000000e+00, 1.00000000e+00, 1.00000000e+00, -1.69779140e+00, 8.26021017e-01 }
+  std::vector<std::vector<double>> sos_{
+    { 9.16782507e-09, 1.83356501e-08, 9.16782507e-09, 1.00000000e+00, -1.82520938e+00, 8.33345838e-01 },
```
RutgerVanBeek, line 101: Why have these numbers changed, is it because of tuning? If so they should be put in a config file I think.

RutgerVanBeek (changes requested):
One small comment, love the `applyButter`function name

march_joint_inertia_controller/CMakeLists.txt
RPKrijnen: I implemented Rutgers suggestion to use the std::clamp function. For the compiler to find this function I needed to upgrade to c++17 according to some stackoverflow page that I can't find anymore.

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
RPKrijnen: These numbers changed due to the update frequency with which we acquire them. It produces a different filter when the input frequency is different.

march_joint_inertia_controller/CMakeLists.txt
Olavhaasie: Yes [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp) is indeed c++17. However, I am not sure if everything will be working correctly, since everything else is c++14. Perhaps change it back to c++14 just to be sure if it's only for this method that can also be written with `std::min` and `std::max`?

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_trajectory_inertia_controller.h
RPKrijnen: Oeh, that's a good suggestion, I'll start on that ASAP.

march_joint_inertia_controller/CMakeLists.txt
RPKrijnen: cool cool, I don't think changing this back affects the performance, so I think I ´ll change this back

march_joint_inertia_controller/include/march_joint_inertia_controller/joint_inertia_controller.h
RPKrijnen: Must have glossed over these haha mb

RutgerVanBeek (dismissed):

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -56,9 +55,10 @@ InertiaEstimator::InertiaEstimator(double lambda, size_t acc_size)
   filtered_joint_torque_.resize(fil_tor_size_, 0.0);
 }
 
-double InertiaEstimator::getAcceleration(unsigned int index)
+double InertiaEstimator::getAcceleration()
 {
-  return acceleration_array_[index];
+  auto a = *acceleration_array_.begin();
+  return a;
```
Olavhaasie: Don't go reinventing the wheel :wink: 
```suggestion
  return acceleration_array_.front();
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -7,23 +7,22 @@
...
 
-double mean(const std::vector<double>& a)
+double mean(const std::list<double>& a)
 {
   double sum = 0.0;
-  for (const double x : a)
+  for (auto it = a.begin(); it != a.end(); it++)
   {
-    sum += x;
+    sum += *it;
```
Olavhaasie: This change was not necessary. You can use the ranged for loop on lists as well

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -137,30 +143,34 @@ void InertiaEstimator::inertiaEstimate()
   correlationCalculation();
   K_i_ = gainCalculation();
   K_a_ = alphaCalculation();
-  const double torque_e = filtered_joint_torque_[0] - filtered_joint_torque_[1];
-  const double acc_e = filtered_acceleration_array_[0] - filtered_acceleration_array_[1];
+  auto ita = filtered_acceleration_array_.begin();
+  auto itt = filtered_joint_torque_.begin();
+  const double torque_e = *itt - *(++itt);
+  const double acc_e = *ita - *(++ita);
```
Olavhaasie, line 145: Is a bit more readable in my opinion and since you don't use it afterwards
```suggestion
  const double torque_e = *itt - *(itt + 1);
  const double acc_e = *ita - *(ita + 1);
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -184,17 +194,19 @@ double InertiaEstimator::vibrationCalculation()
 // Calculate a discrete derivative of the speed measurements
 double InertiaEstimator::discreteSpeedDerivative(const ros::Duration& period)
 {
-  return (velocity_array_[0] - velocity_array_[1]) / period.toSec();
+  auto it = velocity_array_.begin();
+  return (*it - *(++it)) / period.toSec();
```
Olavhaasie, line 194:
```suggestion
  return (*it - *(it + 1)) / period.toSec();
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -137,30 +143,34 @@ void InertiaEstimator::inertiaEstimate()
...
+  return vibration_;
 }
 
 // Calculate the inertia gain for the inertia estimate
 double InertiaEstimator::gainCalculation()
 {
-  return (corr_coeff_ * (filtered_acceleration_array_[0] - filtered_acceleration_array_[1])) /
-         (lambda_ + corr_coeff_ * pow(filtered_acceleration_array_[0] - filtered_acceleration_array_[1], 2));
+  auto it = filtered_acceleration_array_.begin();
+  auto error = *it - *(++it);
```
Olavhaasie, line 161:
```suggestion
  auto error = *it - *(it + 1);
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -137,30 +143,34 @@ void InertiaEstimator::inertiaEstimate()
...
+  return (corr_coeff_ * error) / (lambda_ + corr_coeff_ * pow(error, 2));
 }
 
 // Calculate the correlation coefficient of the acceleration buffer
 void InertiaEstimator::correlationCalculation()
 {
-  corr_coeff_ =
-      corr_coeff_ / (lambda_ + corr_coeff_ * pow(filtered_acceleration_array_[0] - filtered_acceleration_array_[1], 2));
+  auto it = filtered_acceleration_array_.begin();
+  corr_coeff_ = corr_coeff_ / (lambda_ + corr_coeff_ * pow(*it - *(++it), 2));
```
Olavhaasie, line 169:
```suggestion
  corr_coeff_ = corr_coeff_ / (lambda_ + corr_coeff_ * pow(*it - *(it + 1), 2));
```

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -184,17 +194,19 @@ double InertiaEstimator::vibrationCalculation()
...
 {
   // Setup the initial value for the correlation coefficient 100*standarddeviation(acceleration)^2
   double mean_value = mean(standard_deviation);
   double sum = 0;
-  for (const auto& deviation : standard_deviation)
+
+  for (auto it = standard_deviation.begin(); it != standard_deviation.end(); it++)
   {
-    sum += std::pow(deviation - mean_value, 2);
+    sum += std::pow(*it - mean_value, 2);
```
Olavhaasie: This change also wasn't necessary, because ranged for exists for list

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,213 @@
...
+{
+  lambda_ = lambda;
+}
+void InertiaEstimator::setAccSize(size_t acc_size)
+{
+  acc_size_ = acc_size;
+
+  velocity_array_.resize(acc_size_, 0.0);
+  acceleration_array_.resize(acc_size_, 0.0);
+  filtered_acceleration_array_.resize(acc_size_, 0.0);
```
Olavhaasie, line 73: I don't know if you are aware of this but this does not reset the entire list to `0.0`. Only if new values need to be added, when the list is too small, will new values be created. Might not be a problem for you :shrug:

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,213 @@
...
+{
+  // Apply a sixth order Butterworth filter over the effort and acceleration signals
+  auto it1 = z1a.begin();
+  auto it2 = z2a.begin();
+  double x;
+  double x_n;
+
+  for (const auto& so : sos_)
+  {
+    x_n = *acceleration_array_.begin();
```
Olavhaasie:
```suggestion
    x_n = acceleration_array_.front();
```

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,140 @@
...
+  };
+
+  // z1 and z2 vectors are containers for previous time-step values in the butterworth filter
+  // Might be neater to design a separate class for the filter application for arbitrary filter sizes, datasets and
+  // number of signals such as: https://github.com/scipy/scipy/blob/v1.5.2/scipy/signal/_sosfilt.pyx
+  std::list<double> z1a;
+  std::list<double> z2a;
+
+  std::list<double> z1t;
+  std::list<double> z2t;
```
Olavhaasie: I am just wondering, why did you also make these lists? You initialize them once and then always access them in order. I think this would be easier and more efficient as vectors, since they store values sequentially in memory, a vector is almost always more efficient to use. Except for the case I pointed out before, where you used the lists correctly for.

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,213 @@
...
+  return b;
+}
+
+double mean(const std::list<double>& a)
+{
+  double sum = 0.0;
+  for (auto it = a.begin(); it != a.end(); it++)
+  {
+    sum += *it;
+  }
```
Olavhaasie: This can be a ranged for loop

march_joint_inertia_controller/src/inertia_estimator.cpp
```diff
@@ -0,0 +1,213 @@
...
+void InertiaEstimator::initP(unsigned int samples)
+{
+  // Setup the initial value for the correlation coefficient 100*standarddeviation(acceleration)^2
+  double mean_value = mean(standard_deviation);
+  double sum = 0;
+
+  for (auto it = standard_deviation.begin(); it != standard_deviation.end(); it++)
+  {
+    sum += std::pow(*it - mean_value, 2);
+  }
```
Olavhaasie: You can use a ranged for loop again here

march_joint_inertia_controller/src/inertia_estimator.cpp
Olavhaasie: I don't know why you still haven't removed this, but just a friendly reminder :wink:

march_joint_inertia_controller/include/march_joint_inertia_controller/inertia_estimator.h
```diff
@@ -0,0 +1,140 @@
...
+{
+public:
+  /**
+   * \brief Default constructor
+   */
+  InertiaEstimator(double lambda = 0.96, size_t acc_size = 12);
+
+  double getAcceleration();
+  double getJointInertia();
+  double getJointVibration();
```
Olavhaasie: This method is never used?

Olavhaasie (changes requested):
Very well done. It's good to see you have made a lot improvements.

march_joint_inertia_controller/src/inertia_estimator.cpp
RPKrijnen: I'm guessing that, as this is only used during initialization from lists without size, this won't be a problem.

RPKrijnen:
I have changed @Olavhaasie s commits back. i.e. from *(it + 1) back to *(it++), because the suggested method did not work :(

march_joint_inertia_controller/config/joint_inertia_controller.yaml
```diff
@@ -1,12 +1,94 @@
...
+            control_points:
+              inertia:
+                - 0.0
+                - 1.0
+              p:
+                - 150
+                - 300
+              d:
+                - 0
+                - 5
```
Olavhaasie: Indentation is important in YAML files, you have moved the `control_points` key one level deeper than the commit before, which now is on the same level as the array, I am guessing that might not work.

march_joint_inertia_controller/config/joint_inertia_controller.yaml
```diff
@@ -1,12 +1,94 @@
...
```
Olavhaasie: Same goes for these

Olavhaasie (changes requested):

march_joint_inertia_controller/config/joint_inertia_controller.yaml
RPKrijnen: oh damn, totally glossed over this. It is supposed to be a level lower. I'll fix it right away

Olavhaasie (approved):

RutgerVanBeek (approved):

PR #280, opened 2020-07-22
Replace spawner with controller_manger spawn

Olavhaasie:
## Changes
In PM-529, I discovered that the way we spawn and unload controllers was not the best way. The controller spawner we use in the launch file, loads, starts and finally tries to unload them when the spawner is shutdown. However, the controller manager always shuts down faster than the controller spawner can even try to unload them. That describes the warning that says the controller spawner could not unload the controllers. I have now replaced the spawner with just a launcher script, which loads, starts and then forgets about them.

### DIFF
--- a/march_hardware_interface/launch/hardware.launch
+++ b/march_hardware_interface/launch/hardware.launch
@@ -9,8 +9,9 @@
     <group ns="march">
         <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher"/>
 
-        <node name="controller_spawner" pkg="controller_manager" type="spawner" respawn="true"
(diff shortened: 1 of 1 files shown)


Joris-JW (approved):
so small but yet so important, very nice :fire:

RutgerVanBeek (approved):
So you think this is what went wrong when with the spasmes, when shutting down?

Olavhaasie:
> So you think this is what went wrong when with the spasmes, when shutting down?

It could be. Shutting the controller down, while the hardware interface is still operating will cause the effort to stay at the last commanded effort, which is potentially dangerous. So making sure only the hardware interface can shut down the controller is a safe option to do anyways.

PR #282, opened 2020-07-28
Fix; ethercat master watchdog

Joris-JW:
## Description
With the march software team we concluded that the some of the weird exoskeleton behavior is related to connection loss with the IMC slaves. In the code we monitor the slave connection but we do not have any follow up steps if the connection is lost for a longer time. Therefore this PR merges a new feature; the slave Ether-cat watchdog.

It is a new watchdog timer which monitors how long any slave has been lost with the master. If this timer elapses the defined watchdog timer the master raises a new defined error.

## Changes
* Ether-cat watchdog variable
* Added new slave watchdog error
* Raise watchdog error when the timeout is elapsed

## Testing
This has to be tested on the exoskeleton or on the test joint. (preferable the test joint.). Run the hardware interface and remove an Ethernet cable. The program should raise an error after the defined timer in the March4.yaml


march_hardware/include/march_hardware/error/error_type.h
```diff
@@ -31,6 +31,7 @@ enum class ErrorType
   MISSING_REQUIRED_KEY = 119,
   INIT_URDF_FAILED = 120,
   INVALID_SW_STRING = 121,
+  SLAVE_LOST_TIMOUT = 122,
```
Olavhaasie, line 34: Could you also add an error description in project-march/tutorials, if this is merged? :innocent:

march_hardware/src/ethercat/ethercat_master.cpp
```diff
@@ -185,6 +189,16 @@ void EthercatMaster::ethercatLoop()
...
       not_achieved_count = 0;
     }
+
+    const auto delta_t = std::chrono::high_resolution_clock::now() - this->valid_slaves_timestamp_ms_;
+    const auto slave_lost_duration = std::chrono::duration_cast<std::chrono::milliseconds>(delta_t);
+    const std::chrono::milliseconds slave_watchdog_timeout(this->slave_watchdog_timeout_);
+
+    if (slave_lost_duration > slave_watchdog_timeout)
+    {
+      throw error::HardwareException(error::ErrorType::SLAVE_LOST_TIMOUT, "Slave connection lost for %i ms",
```
Olavhaasie: I think there is a problem here. Since `ethercatLoop` runs in its own thread I am not entirely sure how this exception will be caught and if it will be possible to see this exception or not. We might have to use something like [`std::exception_ptr`](https://en.cppreference.com/w/cpp/error/exception_ptr) to propagate the exception across threads. Anyway, it would be important that the ethercat loop stops running and notifies the controller thread that his has happened.

Olavhaasie (changes requested):

march_hardware/src/ethercat/ethercat_master.cpp
Joris-JW: The error was visible after the given time so that is kind off nice but it seemed indeed like the hardware interface did not immediately stopped running. I'll call u to see if we can produce a proper solution.

march_hardware/include/march_hardware/error/error_type.h
Joris-JW: yes!

march_hardware/src/ethercat/ethercat_master.cpp
```diff
@@ -185,6 +189,16 @@ void EthercatMaster::ethercatLoop()
       total_loops = 0;
       not_achieved_count = 0;
     }
+
+    const auto delta_t = std::chrono::high_resolution_clock::now() - this->valid_slaves_timestamp_ms_;
+    const auto slave_lost_duration = std::chrono::duration_cast<std::chrono::milliseconds>(delta_t);
+    const std::chrono::milliseconds slave_watchdog_timeout(this->slave_watchdog_timeout_);
+
+    if (slave_lost_duration > slave_watchdog_timeout)
```
Roelemans, line 204: I think this is temporarily fine, but I think it would be better if the error below contained what slaves were missing. (Ideally you could even keep track from which slave was missing for the most time instead of throwing the slave that was missing at the end of the watchdog timeout)

Roelemans (dismissed):

march_hardware/src/ethercat/ethercat_master.cpp
Joris-JW: Added the last slave which was not detected to the error message.

march_hardware/src/ethercat/ethercat_master.cpp
```diff
@@ -196,8 +203,11 @@ void EthercatMaster::ethercatLoop()
 
     if (slave_lost_duration > slave_watchdog_timeout)
     {
-      throw error::HardwareException(error::ErrorType::SLAVE_LOST_TIMOUT, "Slave connection lost for %i ms",
-                                     this->slave_watchdog_timeout_);
+      this->last_exception_ = std::make_exception_ptr(error::HardwareException(
+          error::ErrorType::SLAVE_LOST_TIMOUT, "Slave connection lost for %i ms", this->slave_watchdog_timeout_));
```
Roelemans: Why do you need to do it like this instead of just directly throwing the error here?

Roelemans (dismissed):

RutgerVanBeek (dismissed):

march_hardware/src/ethercat/ethercat_master.cpp
Olavhaasie: Since there is no place to catch this, because it is in a separate thread. This enables us to read the exception.

RutgerVanBeek (dismissed):

Olavhaasie (approved):
Adds lots of nice stuff, that we should have done much sooner :+1:

Roelemans (approved):

Roelemans (approved):
:partyparrot:

RutgerVanBeek (approved):

PR #284, opened 2020-08-11
Feature/pm 524 Generalize motor controller

Roelemans:
Closes PM-524
Requires [fix/update-imc-states-msg](https://github.com/project-march/march/tree/fix/update-imc-states-msg) branch in the march repo

## Description

In this PR I generalize the hardware interface for different types of motor controllers. I created an abstract class MotorController that specifies an interface each motor controller needs to implement. I then moved some functionality down from Joint class to the motor controller. Most other changes are the renaming of variables from imc to (motor_)controller.

Currently, the iMotionCube states topic still publishes some imc specific information. I was unsure how I could generalize this, so I decided to postpone this to a different PR. Depending on what exactly the format of the corresponding information for other motor controllers is, I might try to generalize the usage of some of its fields. Alternatively, I could just add more fields for each of the unique states of a motor controller type, or create a different topic for each type. Let me know if you have any ideas.

## Changes

* Created abstract MotorController class that is implemented by IMotionCube class.
* Moved some functionality down to IMotionCube
* Renamed lots of variables to (motor_)controller 
* Check joint for existence of motor controller once during startup and remove all runtime checks

## Testing

Imma test on exo tomorrow


march_hardware/include/march_hardware/imotioncube/motor_controller_state.h
```diff
@@ -0,0 +1,35 @@
...
+  double incrementalVelocity;
+
+  std::string statusWord;
+  std::string motionError;
+  std::string detailedError;
+  std::string secondDetailedError;
+  IMCState state;
+  std::string detailedErrorDescription;
+  std::string motionErrorDescription;
+  std::string secondDetailedErrorDescription;
```
RutgerVanBeek: These are imc specific, so shouldn't they be moved to the imc_state.

march_hardware/include/march_hardware/joint.h
```diff
@@ -64,16 +64,17 @@ class Joint
...
   std::string getName() const;
   int getTemperatureGESSlaveIndex() const;
-  int getIMotionCubeSlaveIndex() const;
+  int getMotorControllerSlaveIndex() const;
   int getNetNumber() const;
 
   ActuationMode getActuationMode() const;
 
-  bool hasIMotionCube() const;
+  bool hasMotorController() const;
```
RutgerVanBeek, line 69: May be you could return the typ and `nullptr` if there is none.

march_hardware/src/ethercat/ethercat_master.cpp
```diff
@@ -95,9 +95,9 @@ bool EthercatMaster::ethercatSlaveInitiation(std::vector<Joint>& joints)
 
   for (Joint& joint : joints)
   {
-    if (joint.hasIMotionCube())
+    if (joint.getMotorControllerSlaveIndex() > -1)
```
RutgerVanBeek:
```suggestion
    if (joint.getMotorControllerSlaveIndex() <= 0)
```

RutgerVanBeek (changes requested):
I wander what really in what sense the motor controller is more general then the IMC object. It seems that the MotorController object is just a IMC object where everything is virtual. I cannot imagine that other motor controllers are so similar.

march_hardware/include/march_hardware/imotioncube/motor_controller_state.h
Roelemans: imc_state is basically just an enumeration of "Not Ready to swtich on" and alike states. Hence this is not suitable to move these to that object. The naming is confusing so I improved it slightly by renaming MotorControllerState to MotorControllerStates

march_hardware/src/ethercat/ethercat_master.cpp
Roelemans: No? This should only happen when for ethercat motor controllers, so any motor controller with a slave index > -1. (-1 means no motor controller, and I think Imma use -2 for a non-ethercat motor controller. Or just -1 as well)

march_hardware/include/march_hardware/joint.h
Roelemans: I considered something like that, but it didn't seem very useful anywhere in the code, so I thought hasMotorController would suffice just as well (and it is easier to implement ;))

Roelemans:
> I wander what really in what sense the motor controller is more general then the IMC object. It seems that the MotorController object is just a IMC object where everything is virtual. I cannot imagine that other motor controllers are so similar.

Lemme explain why I think (most of) the methods of the MotorController class are necessary for any motor controller:

All basic necessary functionality is covered by: 
- getters of things like position, speed and current/torque etc. should be measured by any motor controller
- actuation commands for torque/position
- some kind of initialization. For the iMC this is split into initialize and goToOperationEnabled, which might not be necessary for all controllers. These can just have an empty method for one of them. I just realized that I did not have a close enough look at initialize and exactly what roll the EthercatMaster, will look into this tomorrow.
- Perhaps some kind of shutdown (which can be implemented in the destroyer)

That leaves us with the methods
- Every motor controller has some parameters we will want to monitor, hence the getStates method. This might require some additional work to get to work with multiple motor controllers, mainly because one can't subclass ros messages simply (see pr description)
- Virtually every motor controller has multiple actuation modes (position/effort/velocity)
- getSlaveIndex is relevant for any ethercat motorcontroller and will return -2/-1 for any other to indicate ethercat is unused
- checkState is for responding to errors thrown by the motor controller

Does that explain your doubts?

march_hardware/CMakeLists.txt
```diff
@@ -54,6 +54,8 @@ add_library(${PROJECT_NAME}
     include/${PROJECT_NAME}/ethercat/sdo_interface.h
     include/${PROJECT_NAME}/ethercat/slave.h
     include/${PROJECT_NAME}/imotioncube/actuation_mode.h
+    include/${PROJECT_NAME}/imotioncube/motor_controller.h
+    include/${PROJECT_NAME}/imotioncube/motor_controller_state.h
```
Olavhaasie: Why add the "general" motor controller to the `imotioncube` module? Perhaps create a motor controller module which contains the imotioncube module?

march_hardware/include/march_hardware/imotioncube/motor_controller.h
```diff
@@ -0,0 +1,39 @@
...
+public:
+  virtual double getAngleRadAbsolute() = 0;
+  virtual double getAngleRadIncremental() = 0;
+  virtual double getVelocityRadAbsolute() = 0;
+  virtual double getVelocityRadIncremental() = 0;
+  virtual int16_t getTorque() = 0;
+  virtual MotorControllerStates getStates() = 0;
+
+  virtual ActuationMode getActuationMode() const = 0;
+  virtual uint16_t getSlaveIndex() const = 0;
```
Olavhaasie: What if a motor controller isn't a slave or doesn't use the master-slave structure.

march_hardware/include/march_hardware/imotioncube/motor_controller.h
```diff
@@ -0,0 +1,39 @@
...
+
+  virtual float getMotorCurrent() = 0;
+  virtual float getMotorControllerVoltage() = 0;
+  virtual float getMotorVoltage() = 0;
+  virtual bool getIncrementalMorePrecise() const = 0;
+
+  virtual void actuateRad(double target_rad) = 0;
+  virtual void actuateTorque(int16_t target_torque) = 0;
+
+  virtual void goToOperationEnabled() = 0;
```
Olavhaasie: This method is a bit specific (for an interface), maybe rename it to `enable`, which is more general.

march_hardware/include/march_hardware/imotioncube/motor_controller.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace march
+{
+class MotorController
+{
+public:
+  virtual double getAngleRadAbsolute() = 0;
+  virtual double getAngleRadIncremental() = 0;
+  virtual double getVelocityRadAbsolute() = 0;
+  virtual double getVelocityRadIncremental() = 0;
```
Olavhaasie: What if my motor controller does not have either an absolute or incremental encoder?

march_hardware/include/march_hardware/imotioncube/motor_controller.h
```diff
@@ -0,0 +1,39 @@
...
+  virtual int16_t getTorque() = 0;
+  virtual MotorControllerStates getStates() = 0;
+
+  virtual ActuationMode getActuationMode() const = 0;
+  virtual uint16_t getSlaveIndex() const = 0;
+
+  virtual float getMotorCurrent() = 0;
+  virtual float getMotorControllerVoltage() = 0;
+  virtual float getMotorVoltage() = 0;
+  virtual bool getIncrementalMorePrecise() const = 0;
```
Olavhaasie: This also seems like a very specific function that should not be inside the motor controller class. When I read this method call I cannot tell what it would do generally on most controllers.

march_hardware/include/march_hardware/imotioncube/motor_controller.h
```diff
@@ -0,0 +1,39 @@
...
+
+  virtual void actuateRad(double target_rad) = 0;
+  virtual void actuateTorque(int16_t target_torque) = 0;
+
+  virtual void goToOperationEnabled() = 0;
+
+  virtual bool initialize(int cycle_time) = 0;
```
Olavhaasie: What if my motor controller does not need a cycle time, but it does need something else for initialization. That would require adding more parameters and therefore breaking the contract.

march_hardware/include/march_hardware/imotioncube/motor_controller_state.h
Olavhaasie: I still agree with Rutger here, these variables are _very_ IMc specific. You could create Error class where every motor controller could implement its own representation of an error. And then add a method like `getErrorDescription` which returns a universal error description.

march_hardware/include/march_hardware/joint.h
```diff
@@ -139,7 +133,7 @@ class Joint
   double absolute_position_ = 0.0;
   double velocity_ = 0.0;
 
-  std::unique_ptr<IMotionCube> imc_ = nullptr;
+  std::unique_ptr<MotorController> controller_ = nullptr;
```
Olavhaasie: What do we do when a motor controller owns multiple joints? I heard the ODrives control 2 joints? I don't see any of that working in the current implementation.

march_hardware/include/march_hardware/joint.h
```diff
@@ -64,16 +64,17 @@ class Joint
   double getVelocityIUAbsolute();
   double getVelocityIUIncremental();
   float getTemperature();
-  IMotionCubeState getIMotionCubeState();
+  MotorControllerStates getMotorControllerStates();
+  virtual bool checkMotorControllerState(std::ostringstream& error_stream);
```
Olavhaasie: You could build this function into the `MotorControllerStates` object. This separates it from the `Joint` class, which clearly has nothing to do with any of this. Also I would separate the error stream getting from the check method in order to avoid returning variables through references.

march_hardware/src/joint.cpp
```diff
@@ -160,57 +148,12 @@ void Joint::actuateTorque(int16_t target_torque)
...
-
-double Joint::getVelocityIUIncremental()
-{
-  if (!this->hasIMotionCube())
-  {
-    ROS_WARN("[%s] Has no iMotionCube", this->name_.c_str());
-    return -1;
-  }
-  return this->imc_->getVelocityIUIncremental();
+  return this->controller_->getTorque();
```
Olavhaasie, line 143: This will end horribly, if the `controller_` is a `nullptr`, always check you pointers before using them. I now see you do this a lot more, so be warned

march_hardware/test/joint_test.cpp
```diff
@@ -153,16 +153,16 @@ TEST_F(JointTest, GetTemperatureWithoutTemperatureGes)
 
 TEST_F(JointTest, ResetController)
 {
-  EXPECT_CALL(*this->imc, reset(_)).Times(1);
```
Olavhaasie, line 156: I see you are using the IMC here as well as the IMC mock object, you can remove this and just mock the `MotorController` interface

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -106,19 +106,27 @@ march::Joint HardwareBuilder::createJoint(const YAML::Node& joint_config, const
...
+  if (joint_config["imotioncube"])
   {
-    ROS_WARN("Joint %s does not have a configuration for an IMotionCube", joint_name.c_str());
+    controller =
+        HardwareBuilder::createIMotionCube(joint_config["imotioncube"], mode, urdf_joint, pdo_interface, sdo_interface);
+  }
+  //  else if (joint_config["odrive"]) {
+  //      march::ODrive controller =
+  //              HardwareBuilder::createIMotionCube(joint_config["odrive"], mode, urdf_joint);
+  //  }
```
Olavhaasie: Please remove commented out code, it's bad practice

march_hardware/include/march_hardware/imotioncube/imotioncube.h
```diff
@@ -36,42 +37,49 @@ class IMotionCube : public Slave
...
+  virtual float getMotorControllerVoltage() override;
+  virtual float getMotorVoltage() override;
 
+  bool checkState(std::ostringstream& error_msg, std::string joint_name) override;
+  MotorControllerStates getStates() override;
   void setControlWord(uint16_t control_word);
 
-  virtual void actuateRad(double target_rad);
+  virtual void actuateRad(double target_rad) override;
   virtual void actuateTorque(int16_t target_torque);
```
Olavhaasie:
```suggestion
  virtual void actuateTorque(int16_t target_torque) override;
```

Olavhaasie (changes requested):
Overall, I don't see how this solves any problems. This only creates more. I still have to be convinced of the whole idea of a generalized motor controller, since you only use 1 type, where each motor controller works in its own unique way. The way I see this be used is: a new motor controller gets implemented using the interface, but it becomes clear that it misses some functionality/needs to alter some functionality. The interface gets adapted to work with the new motor controller. However, in the process the IMC gets neglected and never used again, since we now have a new one. Ultimately the IMc gets removed and therefore the whole interface is not useful anymore. I see this as a detour in implementing a new motor controller with a lot of extra maintenance.

march_hardware/src/ethercat/ethercat_master.cpp
RutgerVanBeek: Sorry I meant greater or equal then zero, thought it was a bit more intuitive. (but it is a minor detail, so don't bother to much).

march_hardware/include/march_hardware/imotioncube/motor_controller.h
Roelemans: Added a docstring to explain

march_hardware/include/march_hardware/imotioncube/motor_controller.h
Roelemans: I thought about this; the incremental/absolute encoders are a characteristic of Joint rather than of the motor controller, so it would make more sense to move such specifications there. That would require me to subclass Joint into different types, which I preferred to avoid, especially since our current three options for motor controllers will probably be connected to both types of encoders. I think such a refactor of the joint class is better postponed to a moment in the future where an encoder is actually connected to something other than a motor controller.

march_hardware/include/march_hardware/imotioncube/motor_controller.h
Roelemans: Added a docstring with explanation. This method is relevant as long as a motor controller has both an incremental and an absolute encoder.

march_hardware/include/march_hardware/imotioncube/motor_controller.h
Roelemans: Hmm on second thought, I could move the calculations using both encoders downstream to the motor controller. Then these would simplify to getPosition() and getVelocity(). Downside would be that these calculations would be that this code would probably be duplicate in each different MotorController class, but that might again be resolvable by inheriting from a MultipleEncodersMotorControllers class that implements these methods common for motor controllers with both encoders.

That was a nice spam of hersenspinsels, let me know if it makes any sense and what you think.

march_hardware/include/march_hardware/joint.h
Roelemans: We designed the ODrive implementation with this in mind: every Joint is linked to an ODrive object which contains a reference to an ODriveEndpoint object. There is one unique ODrive object for every joint, but only one ODriveEndpoint object for every ODrive. Hence, each ODriveEndpoint object is linked to two ODrive objects. The ODriveEndpoint object handles everything related to communication and the like, similar to the EthercatMaster for our current implementation (except that there are 4 instead of 1).

march_hardware/src/joint.cpp
Roelemans: I build a check in hardware builder to ensure that every joint has a controller at startup, removing the need to check this every time during runtime. 

PS previously, the check at startup was not there, while not everyone of the methods in joint did the nullptr check (e.g. Joint.getIMotionCubeState), so this should actually be an improvement

march_hardware/include/march_hardware/imotioncube/motor_controller.h
Roelemans: Renamed it to prepareActuation, consistent with the naming in the Joint class

march_hardware/include/march_hardware/imotioncube/motor_controller_state.h
Roelemans: Nice I like that suggestion :+1:. As the checkState method basically already implemented such a universal error description, I moved that to a new method getErrorStatus instead of implementing a separate class for this. Let me know what you think.

march_hardware/include/march_hardware/joint.h
Roelemans: Split checkMotorControllerState to check and getErrorStatus, and moved both methods to abstract methods in MotorControllerStates. These are then implemented by a child struct IMotionCubeStates

march_hardware/include/march_hardware/imotioncube/motor_controller.h
Roelemans: Decided that since the 3 motor controllers currently under consideration all have both an incremental and an absolute encoder, this change is not very useful for now. It won't be difficult to implement later on when such a motor controller is actually put to use.

march_hardware/include/march_hardware/imotioncube/motor_controller.h
Roelemans: I think you could alternatively pass something like that to the constructor as well, but I guess that wouldn't be very need since you need it nowhere else. I think I am going to :ostrich: this problem since it is highly dependent on how exactly the startup of another motor controller would work, which is a bit difficult to predict. I will see if I get useful enough insight from the ODrive and the Ingenia to make general statements about this. Otherwise, I think I will leave it up to future generations to redesign this a bit.

march_hardware/include/march_hardware/motor_controller/imotioncube/imotioncube_states.h
```diff
@@ -0,0 +1,45 @@
...
+  uint16_t statusWord;
+  std::string motionError;
+  std::string detailedError;
+  std::string secondDetailedError;
+  IMCStateOfOperation state;
+  std::string detailedErrorDescription;
+  std::string motionErrorDescription;
+  std::string secondDetailedErrorDescription;
+
+  virtual bool checkState()
```
Olavhaasie: This method overrides the one from `MotorControllerStates`, therefore, it should add `override` and not `virtual`
```suggestion
  bool checkState() override
```

march_hardware/include/march_hardware/motor_controller/imotioncube/imotioncube_states.h
```diff
@@ -0,0 +1,45 @@
...
+  {
+    return !(this->state == march::IMCStateOfOperation::FAULT);
+  }
+
+  virtual std::string getErrorStatus()
```
Olavhaasie:
```suggestion
  std::string getErrorStatus() override
```

march_hardware/include/march_hardware/motor_controller/motor_controller_states.h
```diff
@@ -0,0 +1,36 @@
...
+public:
+  MotorControllerStates() = default;
+
+  float motorCurrent;
+  float controllerVoltage;
+  float motorVoltage;
+  int absoluteEncoderValue;
+  int incrementalEncoderValue;
+  double absoluteVelocity;
+  double incrementalVelocity;
```
Olavhaasie: If you are changing this anyway, could you fix the naming of these members to lower_case?

march_hardware/include/march_hardware/motor_controller/imotioncube/imotioncube_states.h
```diff
@@ -0,0 +1,45 @@
...
+  IMotionCubeStates() = default;
+
+  uint16_t statusWord;
+  std::string motionError;
+  std::string detailedError;
+  std::string secondDetailedError;
+  IMCStateOfOperation state;
+  std::string detailedErrorDescription;
+  std::string motionErrorDescription;
+  std::string secondDetailedErrorDescription;
```
Olavhaasie: Could you please fix this naming to lower_case?

march_hardware/src/ethercat/ethercat_master.cpp
```diff
@@ -95,9 +95,9 @@ bool EthercatMaster::ethercatSlaveInitiation(std::vector<Joint>& joints)
 
   for (Joint& joint : joints)
   {
-    if (joint.hasIMotionCube())
+    if (joint.getMotorControllerSlaveIndex() >= 0)
```
Olavhaasie: Actually, a slave index of 0 does not exist, since 0 is the ethercat master
```suggestion
    if (joint.getMotorControllerSlaveIndex() > 0)
```

march_hardware/include/march_hardware/motor_controller/imotioncube/imotioncube_states.h
```diff
@@ -0,0 +1,45 @@
...
+
+  virtual bool checkState()
+  {
+    return !(this->state == march::IMCStateOfOperation::FAULT);
+  }
+
+  virtual std::string getErrorStatus()
+  {
+    std::ostringstream error_stream;
+    std::string state = IMCStateOfOperation(this->statusWord).getString().c_str();
```
Olavhaasie: `c_str()` is not necessary, since it returns a C-like string but you are using `std::string`
```suggestion
    std::string state = IMCStateOfOperation(this->statusWord).getString();
```

march_hardware/src/imotioncube/imotioncube.cpp
```diff
@@ -319,6 +335,38 @@ void IMotionCube::setControlWord(uint16_t control_word)
   this->write16(this->mosi_byte_offsets_.at(IMCObjectName::ControlWord), control_word_ui);
 }
 
+MotorControllerStates& IMotionCube::getStates()
```
Olavhaasie: You shouldn't return references of local variables. I see you tried to fix this using a `static` variable, however, that is not the best solution. Polymorphism should be done using smart pointers like `std::unique_ptr` or `std::shared_ptr`

march_hardware/include/march_hardware/march_robot.h
```diff
@@ -43,23 +43,23 @@ class MarchRobot
...
+  bool isCommunicationOperational();
 
-  std::exception_ptr getLastEthercatException() const noexcept;
+  std::exception_ptr getLastCommunicationException() const noexcept;
 
-  void waitForPdo();
+  void waitForUpdate();
 
-  int getEthercatCycleTime() const;
+  int getCycleTime() const;
```
Olavhaasie: It's nice that you changed these names, however, that does not make them compatible with other motor controllers. The reset and start functions are implemented specifically for the IMotionCube, so you still have to change those methods.

march_hardware/src/march_robot.cpp
```diff
@@ -42,42 +42,42 @@ void MarchRobot::startEtherCAT(bool reset_imc)
...
+    ROS_WARN("Trying to start Communication while it is already active.");
     return;
   }
 
   bool sw_reset = ethercatMaster.start(this->jointList);
 
-  if (reset_imc || sw_reset)
+  if (reset_motor_controllers || sw_reset)
   {
-    ROS_DEBUG("Resetting all IMotionCubes due to either: reset arg: %d or downloading of .sw fie: %d", reset_imc,
```
Olavhaasie, line 55: Again, nice that you changed these names, however, this makes everything so vague. You name it a motor controller, but in the same sentence use the `.sw` files, which are specific for the IMotionCube.

march_hardware/src/joint.cpp
Olavhaasie: What if the hardware is not initialized by the hardware builder, what if it contains a bug that results in creating a `nullptr`? You can't assume that a check somewhere else will protect this class. I know not every method contained the check, but that is not an argument for completely removing it.

Olavhaasie (changes requested):
I see you also started renaming things in the `MarchRobot` class, but not actually changing anything. You are actually making everything more vague and abstract, which is not something we need for software specifically written for one type of hardware. Again, I am still not convinced. To me it seems like you renamed everything from `IMotionCube` to `MotorController` plus a few other fixes.

march_hardware/src/joint.cpp
Roelemans: Hmm do you think a check in the constructor suffice? That will also immediately cover for all the plebs that appearantly like to make methods without these checks. Also, it will make the code less bloated

march_hardware/src/imotioncube/imotioncube.cpp
Roelemans: Aaahhh that makes sense. Fixed :+1:

march_hardware/include/march_hardware/march_robot.h
Roelemans: I was already hesitant about the removal of ethercat from these namings, because that indeed introduces a lot more vagueness. Hence I reverted them for now. 

I do not really understand what you say about reset and start functions; the reset of the MarchRobot directly calls the reset of the MotorController, which should work for any controller. Similarly, the start method calls start for the ethercatMaster, which in turn simply calls initialize for each MotorController. This should work for any Ethercat motor controller. For a none-ethercat motor controller, a new communication master object should be created, with another implementation of the start method.

Ishadijcks:
@TimBuckers @martijnvandermarel Here's an interesting issue

march_hardware/src/march_robot.cpp
Roelemans: A good spot. I changed ".sw file" to "configuration file". Admittedly, that is a bit more vague, but still not to difficult to figure out if you know the system (and this is actually better for noobs that don't know what a .sw file is).

march_hardware/include/march_hardware/ethercat/slave.h
```diff
@@ -31,7 +31,7 @@ class Slave : public PdoSlaveInterface
 
   ~Slave() noexcept override = default;
 
-  uint16_t getSlaveIndex() const
+  int getSlaveIndex() const
```
ThijsRay, line 34: Could you explain the reasoning behind this change?

ThijsRay: Could you explain the reasoning behind this change?

march_hardware/include/march_hardware/joint.h
```diff
@@ -139,8 +89,8 @@ class Joint
   double absolute_position_ = 0.0;
   double velocity_ = 0.0;
 
-  std::unique_ptr<IMotionCube> imc_ = nullptr;
-  std::unique_ptr<TemperatureGES> temperature_ges_ = nullptr;
+  std::shared_ptr<MotorController> controller_ = nullptr;
+  std::shared_ptr<TemperatureGES> temperature_ges_ = nullptr;
```
ThijsRay, line 93: Why has this been changed to a `shared_ptr`?

march_hardware/include/march_hardware/ethercat/ethercat_master.h
```diff
@@ -45,12 +45,19 @@ class EthercatMaster
...
+
+  bool hasValidSlaves();
+
   /**
    * Initializes the ethercat train and starts a thread for the loop.
    * @throws HardwareException If not the configured amount of slaves was found
    *                           or they did not all reach operational state
    */
-  bool start(std::vector<Joint>& joints);
+  bool start();
```
ThijsRay, line 61: Could you explain why the `joints` parameter has been removed and what the return value of this function indicates?

march_hardware/include/march_hardware/ethercat/ethercat_master.h
```diff
@@ -68,7 +75,7 @@ class EthercatMaster
   /**
    * Configures the found slaves to operational state.
    */
-  bool ethercatSlaveInitiation(std::vector<Joint>& joints);
+  bool ethercatSlaveInitiation();
```
ThijsRay, line 81: Could you explain why the `joints` parameter has been removed and what the return value of this function indicates?

march_hardware/include/march_hardware/joint.h
```diff
@@ -26,107 +26,57 @@ class Joint
   /**
    * Initializes a Joint with motor controller and without temperature slave.
    */
-  Joint(std::string name, int net_number, bool allow_actuation, std::unique_ptr<IMotionCube> imc);
+  Joint(std::string name, int net_number, bool allow_actuation, std::shared_ptr<MotorController> controller);
```
ThijsRay, line 29: Why has this been changed from a `unique_ptr` to a `shared_ptr`?

march_hardware/src/imotioncube/imotioncube.cpp
```diff
@@ -220,6 +222,11 @@ double IMotionCube::getAngleRadIncremental()
   return this->incremental_encoder_->getAngleRad(*this, this->miso_byte_offsets_.at(IMCObjectName::MotorPosition));
 }
 
+bool IMotionCube::getIncrementalMorePrecise() const
```
ThijsRay, line 225: I don't get the name for this method. Why is it "`MorePrecise`" if it only returns a `bool`?

march_hardware/src/imotioncube/imotioncube.cpp
```diff
@@ -319,6 +326,39 @@ void IMotionCube::setControlWord(uint16_t control_word)
   this->write16(this->mosi_byte_offsets_.at(IMCObjectName::ControlWord), control_word_ui);
 }
 
+std::unique_ptr<MotorControllerStates> IMotionCube::getStates()
+{
+  std::unique_ptr<IMotionCubeStates> states(new IMotionCubeStates);
```
ThijsRay: I would recommend that you use `make_unique` instead of `new`.

march_hardware/src/joint.cpp
```diff
@@ -18,35 +16,24 @@ Joint::Joint(std::string name, int net_number) : name_(std::move(name)), net_num
...
+  , net_number_(net_number)
+  , allow_actuation_(allow_actuation)
+  , controller_(std::move(controller))
 {
 }
 
-Joint::Joint(std::string name, int net_number, bool allow_actuation, std::unique_ptr<IMotionCube> imc,
-             std::unique_ptr<TemperatureGES> temperature_ges)
+Joint::Joint(std::string name, int net_number, bool allow_actuation, std::shared_ptr<MotorController> controller,
+             std::shared_ptr<TemperatureGES> temperature_ges)
```
ThijsRay, line 28: Why has this been changed to a `shared_ptr`?

march_hardware/src/joint.cpp
ThijsRay: The constructor would only check this once. Adding a check here would ensure that it is always a valid pointer.

march_hardware/src/march_robot.cpp
```diff
@@ -14,51 +14,44 @@
...
-  , ethercatMaster(ifName, this->getMaxSlaveIndex(), ecatCycleTime, ecatSlaveTimeout)
+  , ethercatMaster(ifName, slave_list, ecatCycleTime, ecatSlaveTimeout)
   , pdb_(nullptr)
 {
 }
 
 MarchRobot::MarchRobot(::std::vector<Joint> jointList, urdf::Model urdf,
-                       std::unique_ptr<PowerDistributionBoard> powerDistributionBoard, ::std::string ifName,
-                       int ecatCycleTime, int ecatSlaveTimeout)
+                       std::shared_ptr<PowerDistributionBoard> powerDistributionBoard, ::std::string ifName,
```
ThijsRay, line 27: Why has this been changed to a `shared_ptr`?

march_hardware_builder/include/march_hardware_builder/hardware_builder.h
```diff
@@ -66,21 +67,21 @@ class HardwareBuilder
...
-                                                                     march::SdoInterfacePtr sdo_interface);
-  static std::unique_ptr<march::PowerDistributionBoard>
+  std::shared_ptr<march::IMotionCube> createIMotionCube(const YAML::Node& imc_config, march::ActuationMode mode,
+                                                        const urdf::JointConstSharedPtr& urdf_joint,
+                                                        march::PdoInterfacePtr pdo_interface,
+                                                        march::SdoInterfacePtr sdo_interface);
+  std::shared_ptr<march::TemperatureGES> createTemperatureGES(const YAML::Node& temperature_ges_config,
+                                                              march::PdoInterfacePtr pdo_interface,
+                                                              march::SdoInterfacePtr sdo_interface);
+  std::shared_ptr<march::PowerDistributionBoard>
```
ThijsRay, line 84: Why has this been changed to a `shared_ptr`?

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -106,22 +106,25 @@ march::Joint HardwareBuilder::createJoint(const YAML::Node& joint_config, const
...
   if (!ges)
   {
     ROS_WARN("Joint %s does not have a configuration for a TemperatureGes", joint_name.c_str());
   }
-  return { joint_name, net_number, allow_actuation, std::move(imc), std::move(ges) };
+  return { joint_name, net_number, allow_actuation, std::move(controller), std::move(ges) };
 }
 
-std::unique_ptr<march::IMotionCube> HardwareBuilder::createIMotionCube(const YAML::Node& imc_config,
+std::shared_ptr<march::IMotionCube> HardwareBuilder::createIMotionCube(const YAML::Node& imc_config,
```
ThijsRay, line 127: Why has this been changed to a `shared_ptr`?

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -141,10 +144,14 @@ std::unique_ptr<march::IMotionCube> HardwareBuilder::createIMotionCube(const YAM
   std::ifstream imc_setup_data;
   imc_setup_data.open(ros::package::getPath("march_ems_projects").append("/sw_files/" + urdf_joint->name + ".sw"));
   std::string setup = convertSWFileToString(imc_setup_data);
-  return std::make_unique<march::IMotionCube>(
-      march::Slave(slave_index, pdo_interface, sdo_interface),
-      HardwareBuilder::createAbsoluteEncoder(absolute_encoder_config, urdf_joint),
-      HardwareBuilder::createIncrementalEncoder(incremental_encoder_config), setup, mode);
+
+  std::shared_ptr<march::IMotionCube> imc =
+      std::make_unique<march::IMotionCube>(march::Slave(slave_index, pdo_interface, sdo_interface),
```
ThijsRay: Why are you using `make_unique` if you want to create a `shared_ptr`?

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -196,7 +203,7 @@ HardwareBuilder::createIncrementalEncoder(const YAML::Node& incremental_encoder_
   return std::make_unique<march::IncrementalEncoder>(resolution, transmission);
 }
 
-std::unique_ptr<march::TemperatureGES> HardwareBuilder::createTemperatureGES(const YAML::Node& temperature_ges_config,
+std::shared_ptr<march::TemperatureGES> HardwareBuilder::createTemperatureGES(const YAML::Node& temperature_ges_config,
```
ThijsRay, line 206: Why has this been changed to a `shared_ptr`?

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -210,24 +217,36 @@ std::unique_ptr<march::TemperatureGES> HardwareBuilder::createTemperatureGES(con
...
+  for (auto ges : this->ges_list_)
+  {
+    if (ges->getSlaveIndex() == slave_index)
+    {
+      return ges;
+    }
+  }
+
+  std::shared_ptr<march::TemperatureGES> ges =
+      std::make_unique<march::TemperatureGES>(march::Slave(slave_index, pdo_interface, sdo_interface), byte_offset);
```
ThijsRay: Why are you using `make_unique` if you want to create a `shared_ptr`?

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -246,9 +265,12 @@ std::unique_ptr<march::PowerDistributionBoard> HardwareBuilder::createPowerDistr
       boot_shutdown_byte_offsets["masterOk"].as<int>(), boot_shutdown_byte_offsets["shutdown"].as<int>(),
       boot_shutdown_byte_offsets["shutdownAllowed"].as<int>());
 
-  return std::make_unique<march::PowerDistributionBoard>(march::Slave(slave_index, pdo_interface, sdo_interface),
-                                                         net_monitor_offsets, net_driver_offsets,
-                                                         boot_shutdown_offsets);
+  std::shared_ptr<march::PowerDistributionBoard> pdb =
+      std::make_unique<march::PowerDistributionBoard>(march::Slave(slave_index, pdo_interface, sdo_interface),
```
ThijsRay: Why are you using `make_unique` if you want to create a `shared_ptr`?

march_hardware_builder/src/hardware_builder.cpp
```diff
@@ -210,24 +217,36 @@ std::unique_ptr<march::TemperatureGES> HardwareBuilder::createTemperatureGES(con
...
+  std::shared_ptr<march::TemperatureGES> ges =
+      std::make_unique<march::TemperatureGES>(march::Slave(slave_index, pdo_interface, sdo_interface), byte_offset);
+  this->slave_list_.push_back(ges);
+  this->ges_list_.push_back(ges);
+  return ges;
 }
 
-std::unique_ptr<march::PowerDistributionBoard> HardwareBuilder::createPowerDistributionBoard(
-    const YAML::Node& pdb, march::PdoInterfacePtr pdo_interface, march::SdoInterfacePtr sdo_interface)
+std::shared_ptr<march::PowerDistributionBoard> HardwareBuilder::createPowerDistributionBoard(
```
ThijsRay, line 235: Why has this been changed to a `shared_ptr`?

ThijsRay (commented):
I support the idea that the implementation motor controllers should be more generic, because this would make it easier to implement new motor controllers. However, I am not sure if this is the best interface. Right now, it seems to me that the added layer of abstraction is created with just the iMOTIONCUBE in mind. Other motor controllers might not be able to support all the functionalities that are "required" to implement the `MotorController` class. It might be beneficial to implement other motor controllers before deciding which methods and functionalities are shared between these implementations. This would avoid the creation of an abstraction that is too specific. Such an abstraction would lose its purpose.

march_hardware/include/march_hardware/ethercat/slave.h
Roelemans: Already fixed ;) 
(originally the parallel of this method in MotorController would return -1 for controllers that are not ethercat slaves, but I refactored the ethercat master so that that is no longer necessary)

march_hardware/src/imotioncube/imotioncube.cpp
Roelemans: See the docstring in motor_controller.h. Maybe I should change it to isIncrementalMorePrecise or isIncrementalMorePreciseThanAbsolute (seems a bit long :sweat_smile:)? I am open for other suggestions

march_hardware/include/march_hardware/joint.h
Roelemans: Because I now pass a list of ethercat slaves to ethercat_master (so that march_robot does not need to know which joints are ethercat slaves, and because it makes more sense anyway). So now there are two or three pointers to a a Ges, on in the EthercatMaster, and one in one or two of the joints (previously, the two joints had two separate ges objects for the same ges).

march_hardware/include/march_hardware/joint.h
Roelemans: Same reason for the ges, there is now one pointer in EthercatMaster, and one in the MarchRobot

march_hardware/src/joint.cpp
Roelemans: See comments on the others

march_hardware/src/march_robot.cpp
Roelemans: Same for motor controller

march_hardware_builder/include/march_hardware_builder/hardware_builder.h
Roelemans: Same

march_hardware_builder/src/hardware_builder.cpp
Roelemans: Same

march_hardware_builder/src/hardware_builder.cpp
Roelemans: Same

march_hardware_builder/src/hardware_builder.cpp
Roelemans: Same

march_hardware/src/imotioncube/imotioncube.cpp
Roelemans: Thanks! I am still a bit of a noob when it comes to smart pointers :sweat_smile:

march_hardware_builder/src/hardware_builder.cpp
Roelemans: Oopsie :sweat_smile:

march_hardware_builder/src/hardware_builder.cpp
Roelemans: Oopsie :sweat_smile:

march_hardware_builder/src/hardware_builder.cpp
Roelemans: Oopsie :sweat_smile:

march_hardware/include/march_hardware/ethercat/ethercat_master.h
Roelemans: Previously, the joints parameter was used by the ethercat master to access the ethercat slaves to call initSdo on each. The EthercatMaster is now passed a list of ethercat slaves at initiation, making joints parameter unnecessary. The return value is now explained in the docstring

march_hardware/include/march_hardware/ethercat/ethercat_master.h
Roelemans: See above

PR #287, opened 2020-08-18
Fix: initialize variables

Olavhaasie:
## Description
When building in release mode, it gives an error here stating that it could be used uninitialized, so I initialized it to 0.0.

## Changes
* Initialize doubles to 0.0


Roelemans (approved):

RPKrijnen (approved):
Cool cool, was thinking of making this PR myself. Thankyou for fixing this ahead of me :D

PR #288, opened 2020-08-21
Feature/add odrive two motors

Roelemans:
## Description

Intgerate the ODrive with the hardware interface according to the interface defined in https://github.com/project-march/hardware-interface/pull/284


Roelemans:
M5 will not integrate this functionality anymore.

